PR #9, opened 2018-07-29
Heapster baseline fix

### LINKED ISSUE #8: Assure that heapster baseline files are named correctly
When creating a heapster baseline the name of the file is assumed. The source filename should be used for consistency.

DavidXArnold:
#### What does this PR do?
- Fixes #8 
- assures that the source and destination heapster baseline files names are consistently used.
- adds a note to the contribution document about manual version

https://github.com/cloudability/metrics-agent/compare/master...DavidXArnold:heapsterBaselineFix?expand=1#diff-5cf9ac41985e0ce94b1270c7ec700302R193

`make deploy-local`

#8 

#### Developer Done List

- [x] Tests Added/Updated
- [x] Updated README.md

- [ ] Tests Pass Locally
- [ ] CI Build Passes

### DIFF
--- a/kubernetes/kubernetes.go
+++ b/kubernetes/kubernetes.go
@@ -191,10 +191,10 @@ func (ka KubeAgentConfig) collectMetrics(
 	defer util.SafeClose(hme.Close, &rerr)
 
 	baselineMetricSample, err := util.MatchOneFile(path.Dir(config.msExportDirectory.Name()), "/baseline-metrics-export*")
-	if err == nil {
+	if err != nil || baselineMetricSample != "" {
 		err = handleBaselineHeapsterMetrics(msd, baselineMetricSample, hme.Name())
 		if err != nil {
-			return err
+			log.Printf("Error updating Heapster Baseline: %v", err)
 		}
 	}
 
--- a/kubernetes/kubernetes.go
+++ b/kubernetes/kubernetes.go
@@ -190,14 +190,9 @@ func (ka KubeAgentConfig) collectMetrics(
 
 	defer util.SafeClose(hme.Close, &rerr)
 
-	baselineMetricSample, err := filepath.Glob(path.Dir(config.msExportDirectory.Name()) + "/baseline-metrics-export*")
-
-	if err != nil {
-		log.Printf("Error encountered reading export directory: %v", err)
-	}
-
-	if len(baselineMetricSample) >= 1 {
-		err = handleBaselineHeapsterMetrics(msd, baselineMetricSample[0], hme.Name())
+	baselineMetricSample, err := util.MatchOneFile(path.Dir(config.msExportDirectory.Name()), "/baseline-metrics-export*")
+	if err == nil {
+		err = handleBaselineHeapsterMetrics(msd, baselineMetricSample, hme.Name())
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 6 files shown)


kubernetes/kubernetes.go
```diff
@@ -186,14 +185,22 @@ func (ka KubeAgentConfig) collectMetrics(
...
 
-	err = handleBaselineHeapsterMetrics(msd, baselineMetricSample, hme.Name())
+	baselineMetricSample, err := filepath.Glob(path.Dir(config.msExportDirectory.Name()) + "/baseline-metrics-export*")
+
 	if err != nil {
-		return err
+		log.Printf("Error encountered reading export directory: %v", err)
+	}
+
+	if len(baselineMetricSample[0]) >= 1 {
```
housejester: could this nil point if the call to file.Glob errored out? we check for that, but only log the error (line 196), and don't return.

housejester: filepath.Glob returns a slice of matches. here we're checking the length of the first match. do we intend to do that? is there a case where the first match is an empty string? or is it meant to be checking the number of matches instead?

kubernetes/kubernetes.go
```diff
@@ -186,14 +185,22 @@ func (ka KubeAgentConfig) collectMetrics(
...
+		err = handleBaselineHeapsterMetrics(msd, baselineMetricSample[0], hme.Name())
```
housejester: this feels weird that we only process the first one. If we assume/only expect to have one, we could clarify that by asserting it. or adding a helper method that could do it for us. eg `baselineMetricSample, err := MatchOneFile("baseline-metrics-export*")`. possibly even `MustMatchOneFile` if we think it should panic if it matches multiple.

kubernetes/kubernetes.go
```diff
@@ -186,14 +185,17 @@ func (ka KubeAgentConfig) collectMetrics(
...
+		return fmt.Errorf("unable to retrieve raw heapster metrics: %s", err)
 	}
 
 	defer util.SafeClose(hme.Close, &rerr)
 
-	err = handleBaselineHeapsterMetrics(msd, baselineMetricSample, hme.Name())
-	if err != nil {
-		return err
+	baselineMetricSample, err := util.MatchOneFile(path.Dir(config.msExportDirectory.Name()), "/baseline-metrics-export*")
+	if err == nil {
```
egs5053: I would refactor this to instead be:

`if err != nil {
  return err
}
err = handleBaselineHeapsterMetrics(msd, baselineMetricSample, hme.Name())
if err != nil {
  return err
}
`

egs5053: unless you want to continue through the rest of the function if it fails.

DavidXArnold: We want it to copy the baseline if one is found else continue through.

kubernetes/kubernetes.go
```diff
@@ -186,14 +185,18 @@ func (ka KubeAgentConfig) collectMetrics(
...
 	}
 
 	defer util.SafeClose(hme.Close, &rerr)
 
-	err = handleBaselineHeapsterMetrics(msd, baselineMetricSample, hme.Name())
-	if err != nil {
-		return err
+	baselineMetricSample, fileCount, err := util.MatchOneFile(
+		path.Dir(config.msExportDirectory.Name()), "/baseline-metrics-export*")
+	if err == nil || fileCount == 0 {
```
egs5053: so if err != nil but fileCount is zero we still want to continue?

DavidXArnold: yes. per your suggestion I refactored for better read-ability

PR #31, opened 2019-01-24
don't launch heapster if no heapster found

mollylogue:
#### What does this PR do?
- If heapster not found, do not launch heapster (and don't error out. collect summaries)
- If no heapster found _and_ no node summaries, should error out and exit

- Run locally. Verify that heapster is not launched and summaries are collected

giphy-1

#### Developer Done List

- [ ] Tests Added/Updated
- [ ] Updated README.md

- [ ] Tests Pass Locally
- [ ] CI Build Passes

### DIFF
--- a/kubernetes/heapster.go
+++ b/kubernetes/heapster.go
@@ -14,10 +14,7 @@ import (
 
 	"github.com/cloudability/metrics-agent/util"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/kubernetes"
-	apiv1 "k8s.io/client-go/pkg/api/v1"
-	appsv1beta1 "k8s.io/client-go/pkg/apis/apps/v1beta1"
 	"k8s.io/client-go/rest"
 )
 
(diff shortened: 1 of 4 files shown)


kubernetes/heapster.go
```diff
@@ -114,221 +111,26 @@ func getHeapsterURL(clientset kubernetes.Interface, clusterHostURL string) (URL
...
 
+	var me heapsterMetricExport
 	if err := json.Unmarshal(*body, &me); err != nil {
 		log.Printf("malformed response from heapster running at: %v", config.HeapsterURL)
-	}
-
-	if !outerTest {
-		log.Printf("Could not connect to heapster at: %v, launching a copy in the Cloudability namespace.",
+	} else if !outerTest {
+		return fmt.Errorf("no heapster found")
```
housejester: seems like this block should go first. and may be able to take out the if/else structures (if the ifs are returning...)

kubernetes/heapster.go
```diff
@@ -114,221 +111,26 @@ func getHeapsterURL(clientset kubernetes.Interface, clusterHostURL string) (URL
...
+	} else if len(me) < 10 {
+		log.Printf("Received empty or malformed response from heapster running at: %v",
 			config.HeapsterURL)
```
housejester: should this be an error return?

kubernetes/heapster.go
```diff
@@ -114,221 +111,26 @@ func getHeapsterURL(clientset kubernetes.Interface, clusterHostURL string) (URL
...
 		client, config.HeapsterURL, http.MethodGet, config.BearerToken, retryCount, true)
 
-	var me heapsterMetricExport
+	if err != nil {
+		return err
+	}
 
+	var me heapsterMetricExport
 	if err := json.Unmarshal(*body, &me); err != nil {
 		log.Printf("malformed response from heapster running at: %v", config.HeapsterURL)
```
housejester: would this be an error return?

kubernetes/kubernetes.go
```diff
@@ -550,13 +553,21 @@ func ensureMetricServicesAvailable(config KubeAgentConfig) KubeAgentConfig {
 		}
 	}
 	if config.CollectHeapsterExport {
-		config, err = ensureValidHeapster(config)
+		client := &config.HTTPClient
```
housejester: could probably inline this variable? only used in the one place

kubernetes/heapster.go
mollylogue: Depends. Do we want this to be a fatal error? If we can connect to heapster but the response is malformed, do we want to give up on connecting to heapster in the future?

housejester: seems like it should. if we're expecting to pull from heapster, but can't prove we're getting good data, it should error. at least from this method. i could see the caller deciding to be ok with it and just log it, if we're also collecting summaries and just "best effort" for heapster.

PR #38, opened 2019-07-02
Log warning when node provider ID doesn't exist

egs5053:
#### What does this PR do?
Logs a warning when the node ProviderID is empy.

`kubernetes/nodecollection.go`

Running the metric agent on a node that does not have a ProviderID and ensuring the log is present.

N.

N/A

#### Developer Done List

- [x] Tests Added/Updated
- [x] Updated README.md

- [ ] Tests Pass Locally
- [ ] CI Build Passes

### DIFF
--- a/kubernetes/nodecollection.go
+++ b/kubernetes/nodecollection.go
@@ -3,6 +3,7 @@ package kubernetes
 import (
 	"crypto/tls"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log"
 	"net/http"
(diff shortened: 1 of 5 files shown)


kubernetes/nodecollection.go
```diff
@@ -81,6 +81,10 @@ func downloadNodeData(prefix string,
 	}
 
 	for _, n := range nodes.Items {
+		if n.Spec.ProviderID == "" {
+			log.Println("Warning: Provider ID for node does not exist")
```
DavidXArnold: Please also log the node name that does not have a providerID in the message as well as something along the lines of "If this condition persists it will cause inconsistent cluster allocation"

DavidXArnold: Should we add a test for this?  Should be able to mock node Spec in a similar fashion to this:
https://github.com/cloudability/metrics-agent/blob/master/kubernetes/nodecollection_test.go#L24-L27

PR #66, opened 2020-01-09
Add node ready checks to nodecollection methods: downloadNodeData and…

finalfreq:
… ensureNodeSource.

#### What does this PR do?
Adds a nodeready status check in the nodecollection module. In one case we want to use the first ready node to find the node source address, and when we are polling node data we want to make sure a node is ready before requesting node summaries, if the node isn't ready we will just skip it. If no nodes were ready during initialization this will result in an error as well as if no nodes were ready when we attempt to get node summaries. 
Make sure the checks are in the right place (I'm not sure if in the downloadNodeData if it should be above the if statement it is currently nested under or not). Make sure logic looks good
Will do a manual test using one of our test clusters. 
#### Developer Done List

- [x] Tests Added/Updated
- [ ] Updated README.md

- [ ] Tests Pass Locally
- [ ] CI Build Passes

### DIFF
--- a/kubernetes/nodecollection.go
+++ b/kubernetes/nodecollection.go
@@ -91,20 +91,20 @@ func downloadNodeData(prefix string,
 				"If this condition persists it will cause inconsistent cluster allocation")
 		}
 
+		// Nodes running in a cluster may not all be ready, check each node individually and if they are not ready
+		// skip the node and create a log.
+		nodeReady := v1Node.IsNodeReady(&n)
+		if !nodeReady {
+			log.Warnf("Node %s was not ready when attempting to get node summary", n.Name)
+			continue
+		} else if anyNodeReady == false {
+			anyNodeReady = true
+		}
+
 		// retrieve node summary directly from node if possible and allowed.
 		// The config shouldn't allow direct connection if Fargate nodes were
 		// found in the cluster at startup, but check again here to be safe.
 		if config.nodeRetrievalMethod == direct && !isFargateNode(n) {
-			// Nodes running in a cluster may not all be ready, check each node individually and if they are not ready
-			// skip the node and create a log.
-			nodeReady := v1Node.IsNodeReady(&n)
-			if !nodeReady {
-				log.Warnf("Node %s was not ready when attempting to get node summary", n.Name)
-				continue
-			} else if anyNodeReady == false {
-				anyNodeReady = true
-			}
-
 			ip, port, err := nodeSource.NodeAddress(&n)
 			if err != nil {
 				return nil, fmt.Errorf("error: %s", err)
@@ -149,6 +149,7 @@ func downloadNodeData(prefix string,
 //ensureNodeSource validates connectivity to the kubelet metrics endpoints.
 // Attempts direct connection to the node summary & container stats endpoint
 // if possible and allowed, otherwise attempts to connect via kube-proxy
+// nolint: gocyclo
 func ensureNodeSource(config KubeAgentConfig) (KubeAgentConfig, error) {
 
 	nodeHTTPClient := http.Client{
--- a/kubernetes/nodecollection.go
+++ b/kubernetes/nodecollection.go
@@ -58,12 +58,16 @@ func (cns ClientsetNodeSource) GetReadyNodes() ([]v1.Node, error) {
 		if nodeReady {
 			readyNodes = append(readyNodes, n)
 		} else {
-			log.Infof("node, %s, is in a notready state", n.Name)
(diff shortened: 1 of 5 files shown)


kubernetes/nodecollection.go
```diff
@@ -81,12 +83,24 @@ func downloadNodeData(prefix string,
...
 				"If this condition persists it will cause inconsistent cluster allocation")
 		}
 
+		// Nodes running in a cluster may not all be ready, check each node individually and if they are not ready
+		// skip the node and create a log.
+		nodeReady := v1Node.IsNodeReady(&n)
+		if !nodeReady {
+			log.Warnf("Node %s was not ready when attempting to get node summary", n.Name)
+			continue
+		} else if anyNodeReady == false {
```
jpepin: Could probably skip this check and just make it true each time. Might help with the cyclomatic complexity lint, too. You could even make it a counter if we wanted to log the number of ready/not ready nodes, but that could be overkill...

kubernetes/nodecollection.go
```diff
@@ -153,7 +172,22 @@ func ensureNodeSource(config KubeAgentConfig) (KubeAgentConfig, error) {
...
+	// Some nodes may not be ready, we want to find the first node that is ready so we can set the node source
+	var firstReadyNode *v1.Node
+
+	for _, n := range nodes.Items {
+		nodeReady := v1Node.IsNodeReady(&n)
+
+		if nodeReady {
+			firstReadyNode = &n
+		}
+	}
```
jpepin: It might be more efficient to break out of the loop on the first ready node found, rather than processing the whole list. Pulling that into a function that just returns a firstReadyNode would also be a drop-in way to lower the cyclomatic complexity.....the most annoying lint 😆

finalfreq: oh yes it is the most annoying lint, feel like sometimes if you fix it you just potentially make the code less readable. I will take a look at this!

kubernetes/nodecollection.go
finalfreq: Ok cool sounds good to me

kubernetes/nodecollection.go
```diff
@@ -145,10 +132,111 @@ func downloadNodeData(prefix string,
...
+	workDir           *os.File
+	ClusterHostURL    string
+	containersRequest []byte
+}
+
+// directNodeFetch retrieves node stats directly from the node api
+func directNodeFetch(nodeSource NodeSource, config KubeAgentConfig, n *v1.Node, nd nodeFetchData) error {
+	ip, port, err := nodeSource.NodeAddress(n)
+	if err != nil {
+		return fmt.Errorf("problem getting node address: %s", err)
```
finalfreq: I noticed that previously this error would have been `return nil, fmt.Errorf` in this case we would just be adding it as a failed node in the failedNodeList. I'm not sure what the intent was originally but i believe this would cause a different outcome. Feel free to correct me if I'm wrong!

jpepin: No, you're correct. I'm hoping someone who knew the intent of that error could weigh in, because it doesn't seem necessary to fail the whole node fetch if that part of the direct node fetch fails. In other cases we just move on to trying proxy, so I'm not sure why that error was being handled differently.

kubernetes/nodecollection.go
```diff
@@ -148,48 +261,38 @@ func ensureNodeSource(config KubeAgentConfig) (KubeAgentConfig, error) {
...
 		return config, fmt.Errorf("error retrieving nodes: %s", err)
 	}
 
-	ip, port, err := clientSetNodeSource.NodeAddress(&nodes.Items[0])
+	firstNode := &nodes[0]
+
+	ip, port, err := clientSetNodeSource.NodeAddress(firstNode)
 	if err != nil {
 		return config, fmt.Errorf("error retrieving node addresses: %s", err)
 	}
```
jpepin: Not going to block this PR, but I just noticed that we're failing this branch on being unable to get the node address, also, despite that info only being used in the direct connection test. Might want to rethink that in future....

kubernetes/nodecollection_test.go
```diff
@@ -252,45 +290,65 @@ func TestDownloadNodeData(t *testing.T) {
...
+			ka,
+			ed,
+			ns,
+		)
+
+		if err == nil {
+			t.Error("Expected no nodes found error")
+		}
+
+		if err.Error() != "cloudability metric agent is unable to get a list of nodes: 0 nodes were ready" {
```
jpepin: If we ever change this error message, this test will fail. But at that point I guess we can just make it a constant.

Issue #83, opened 2020-06-11
Unable to pull image via Helm -- Need to Release


physik932:
Hello!  While utilizing the metrics-agent chart, we were unable to pull the image because the release used in `deployment.yaml` references `image: 'cloudability/metrics-agent:0.1.0`, but this version isn't released on this repository.

https://github.com/cloudability/metrics-agent/blob/master/deploy/charts/metrics-agent/templates/deployment.yaml#L26

If you cut a release for 0.1.0, the image created here should work.  

Thanks for your time!

johnjeffers:
@physik932 The chart's been updated to use `latest` instead of the chart version. This is the same way the Cloudability app works when you generate a template in the app. If you prefer to hardcode the version, you can override it in `values.yaml` https://github.com/cloudability/metrics-agent/pull/84

Issue #82, opened 2020-05-19
Non-numeric user (agent) Error


webbdog28:
This is the error I'm getting when trying to run the metrics-agent in Kubernetes 1.15.10:

"Error: container has runAsNonRoot and image has non-numeric user (agent), cannot verify user is non-root"

This was fixed by changing the deployment.yaml file and adding the following:

      securityContext: 
        runAsUser: 1000

I would suggest either adding that to the deployment.yaml or updating the Dockerfile.

johnjeffers:
https://github.com/cloudability/metrics-agent/pull/84

johnjeffers:
Reopening, realized the Helm chart fix does not address this.

paulbhart:
The better solution is the one I listed here https://github.com/cloudability/metrics-agent/issues/85.  It doesn't change functionality just ensure the USER command in the docker file is a numeric id vs. a alpha value.

mnorbury:
This should be fixed by #87, closing the issue. Please re-open if this continues to be a problem.

PR #84, opened 2020-06-12
multiple updates to the helm chart

johnjeffers:
#### What does this PR do?
Multiple changes to the helm chart
- **BREAKING CHANGE** - `deployment.yaml` now uses `image.name:image.tag` instead of `image.repository:Chart.appversion` (https://github.com/cloudability/metrics-agent/issues/83)
- `image.tag` defaults to `latest` in `values.yaml` to match what is present in the generated template provided by the Cloudability app
- Hardcode `replicaCount` to 1 in `deployment.yaml`, removed the variable from `values.yaml`. Only one copy of the agent should run.
- Bump chart version to `0.2.0`
- Bump app version to `1.2.0`
- Misc template cleanup in `deployment yaml`
- Add more information to `values.yaml` to make the values clearer.
- Remove David Arnold from `CODEOWNERS`
- Add `securityContext` (https://github.com/cloudability/metrics-agent/issues/82)

values.yaml

`helm template cloudability .`

#### Developer Done List

- [ ] Tests Added/Updated
- [ ] Updated README.md

- [ ] Tests Pass Locally
- [ ] CI Build Passes

### DIFF
--- a/deploy/charts/metrics-agent/Chart.yaml
+++ b/deploy/charts/metrics-agent/Chart.yaml
@@ -14,8 +14,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.1.0
+version: 0.2.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
-appVersion: 0.1.0
+appVersion: 1.2.0
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,5 +1,4 @@
 *  @egs5053
-*  @DavidXArnold
 *  @housejester
 *  @mollylogue
 *  @mnorbury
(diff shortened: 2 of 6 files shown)


deploy/charts/metrics-agent/Chart.yaml
```diff
@@ -14,8 +14,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.1.0
+version: 0.2.0
```
mnorbury, line 17: I noticed in the Makefile the Helm version uses the release version. Should it be this version?

https://github.com/cloudability/metrics-agent/blob/bbeb62dc81ce36a65f10a21cdc1d35aeeb3d3d56/Makefile#L58

johnjeffers: Good catch. I took the hardcoded versions out of the `helm package` command in the Makefile. Not sure why we would need that, vs. just sticking with the versions that are in the chart.

PR #91, opened 2020-10-22
Make metric collection retry limit configurable

jpepin:
#### What does this PR do?
Make the collection retry limit configurable. Currently the retry limit is hard-coded to 3 with an added backoff each time, which can cause the agent to spend an excessive amount of time attempting to connect to a node that is no longer in the cluster. This is problematic for very large clusters, or clusters with heavy spot usage.

Users who know they have a highly dynamic cluster may choose to set a lower retry limit to address this issue.

#### Where should the reviewer start?
kubernetes/kubernetes.go:435 and kubernetes/nodecollection.go:265 is where the new retry value is passed to the client. The rest of the changes are just threading this value through.

#### How should this be manually tested?
You can run the agent locally with a flag or env var to alter the retries.

#### Developer Done List

- [x] Tests Added/Updated
- [x] Updated README.md
- [x] Verified backward compatible
- [x] Verified database migrations will not be catastrophic
- [x] Considered Security, Availability and Confidentiality

#### Reviewer/Approval Done List

- [ ] Tests Pass Locally
- [ ] CI Build Passes
- [ ] Verified README.md is updated
- [ ] Verified changes are backward compatible
- [ ] Reviewed impact to Security, Availability and Confidentiality (if issue found, add comments and request changes)

PR #92, opened 2020-10-23
Revert "Make metric collection retry limit configurable"

jpepin:
Reverts cloudability/metrics-agent#91

Our CI builds aren't working, so revert this change until we can troubleshoot that.


rrobb6 (approved):
:shipit:

PR #102, opened 2020-11-30
Snyk scan/monitor GitHub actions step

mollylogue:
#### What does this PR do?
adds a snyk test/monitor GitHub actions step for every commit pushed, in keeping w/ the current build behavior in CircleCI

.github/workflows/config.yml 

Running this action from my forked repository

We will need to generate a new snyk token and add it as a secret in this repository for this action to work

#### Developer Done List

- [ ] Tests Added/Updated
- [ ] Updated README.md

- [ ] Tests Pass Locally
- [ ] CI Build Passes

### DIFF
--- a/.github/workflows/config.yml
+++ b/.github/workflows/config.yml
@@ -4,6 +4,24 @@ on: [push]
 
 jobs:
 
+  snyk_scan:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.14'
+      - uses: snyk/actions/setup@master
+      - uses: actions/checkout@v2
+      - name: Snyk Test
+        run: snyk test  --severity-threshold=medium --org=cloudability
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+      - name: Snyk monitor
+        run: snyk monitor --org=cloudability
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+
+
   build:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/config.yml
```diff
@@ -4,6 +4,24 @@ on: [push]
...
 jobs:
 
+  snyk_scan:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.14'
+      - uses: snyk/actions/setup@master
+      - uses: actions/checkout@v2
```
rrobb6, line 14: Are these `uses` arrays defining when the scan is going to run?

mollylogue: I believe uses just means that these steps need to run before any actions

.github/workflows/config.yml
```diff
@@ -4,6 +4,24 @@ on: [push]
...
+      - name: Snyk Test
+        run: snyk test  --severity-threshold=medium --org=cloudability
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+      - name: Snyk monitor
+        run: snyk monitor --org=cloudability
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
```
mnorbury, line 22: Just to double check, this only runs on the master build right?

mollylogue: No, it runs every commit. I think that's what circleci does currently?

mnorbury: Yeah, looks like you are right. Strange that. I think this means we will register all branches to be monitored going forward, even if they've since been merged in.

mnorbury (approved):
LGTM

.github/workflows/config.yml
mollylogue: I could update to have the test step run on every commit and the monitor just run on the master build? Though maybe we want to maintain legacy behavior from circleci for now.

Issue #90, opened 2020-10-22
Add support for Kubernetes 1.18


giannisbetas:
After upgrading one of the Kubernetes clusters to 1.18 the cloudability-metrics-agent started crashing with the following error:
```
time="2020-10-22T14:33:15Z" level=info msg="Starting Cloudability Kubernetes Metric Agent version: 1.2.2"
time="2020-10-22T14:33:15Z" level=info msg="Primary metrics will be collected from each node."
time="2020-10-22T14:33:16Z" level=warning msg="Warning non-fatal error: Agent error occurred retrieving node source metrics: unable to retrieve node metrics. Please verify RBAC roles: <nil> "
time="2020-10-22T14:33:16Z" level=warning msg="For more information see: https://support.cloudability.com/hc/en-us/articles/360008368193-Kubernetes-Metrics-Agent-Error-Messages"
time="2020-10-22T14:33:16Z" level=fatal msg="unable to retrieve node summaries or heapster export: unable to retrieve node metrics. Please verify RBAC roles: <nil>"
```

In Cloudability there is no option to generate a manifest for Kubernetes 1.18.

jpepin:
As part of their initiative to deprecate cadvisor and streamline core metrics, Kubernetes 1.18 has turned off by default some of the APIs that we use to gather certain key metrics. You can see more details about that here: https://github.com/kubernetes/kubernetes/issues/68522

Cloudability doesn't currently support Kubernetes 1.18 for that reason. You can manually enable the cadvisor endpoints in your cluster, and the metrics-agent will be able to collect and upload data. Because not all users have the ability or desire to do that, we have chosen not to officially support 1.18 until we have a full replacement source, which we anticipate will be ready in early 2021.

giannisbetas:
Thank you @jpepin 
I can confirm that the agent is reporting metrics after setting `enable-cadvisor-json-endpoints` to `true` for all the kubelets.
A couple of notes on this issue:
- I couldn't find any documentation/notes in regards to Kubernetes 1.18 support, perhaps a sections in this repo's README would make sense.
- The error message regarding RBAC is misleading.
- I noticed that the agent was crashing when not all nodes had the `enable-cadvisor-json-endpoints` set to `true`.

jpepin:
@giannisbetas Thanks for letting us know -- we'll clarify the question of version in our docs, and work to disambiguate the documentation for that error as well.
[...]

PR #116, opened 2021-03-09
Encode connection type at the endpoint level rather than the cluster level

jpepin:
#### What does this PR do?
Currently the metrics agent assumes that all node endpoints will be able to use the same type of connection (direct or proxy), and if one endpoint fails on the direct connection attempt, all endpoints fall back to using a proxy connection.

In some clusters the new `metrics/cadvisor` endpoint is not accessible directly, so this PR simplifies the connection logic to set the allowed connection on each endpoint rather than cascading through each connection type for all endpoints.

Take a look at the Connection type and the EndpointMask changes

This has been tested on several clusters with varying permissions configurations

After this change, the use of separate clients for proxy and direct is starting to seem like an unecessary complexity, but in the interest of keeping this PR small I didn't tackle that part.

This change would also enable us to avoid needing to disable direct connection on clusters that have a single Fargate node in them, as we could handle connection type at the node level rather than at the cluster level.

#90 
#### Developer Done List

- [x] Tests Added/Updated
- [x] Updated README.md

- [ ] Tests Pass Locally
- [ ] CI Build Passes

### DIFF
--- a/kubernetes/endpoint_test.go
+++ b/kubernetes/endpoint_test.go
@@ -1,0 +1,162 @@
+package kubernetes
+
+import (
+	"testing"
+)
+
+func TestConnection(t *testing.T) {
+	t.Run("Connection bitmask handles Unreachable correctly", func(t *testing.T) {
+		exampleConnection := Connection(0)
+		if exampleConnection != Unreachable {
+			t.Errorf("zero-value Connection should be unreachable, got %s", exampleConnection)
+		}
+		// make sure no weird shenanigans happen with double Unreachable
+		exampleConnection.ClearMethod(Unreachable)
+		if exampleConnection != Unreachable {
+			t.Errorf("zero-value Connection should be unreachable, got %s", exampleConnection)
+		}
+		// note: Don't do this, use SetUnreachable
+		exampleConnection.AddMethod(Unreachable)
+		if exampleConnection != Unreachable {
+			t.Errorf("zero-value Connection should be unreachable, got %s", exampleConnection)
+		}
+
+		exampleConnection.SetUnreachable()
+		if exampleConnection != Unreachable {
+			t.Errorf("zero-value Connection should be unreachable, got %s", exampleConnection)
+		}
+	})
+	t.Run("Connection bitmask handles Proxy correctly", func(t *testing.T) {
+		exampleConnection := Connection(0)
+		exampleConnection.AddMethod(Proxy)
+
+		if !exampleConnection.hasMethod(Proxy) {
+			t.Errorf("expected Proxy, got %s", exampleConnection)
+		}
+		exampleConnection.ClearMethod(Proxy)
+		if exampleConnection != Unreachable {
+			t.Errorf("cleared connection should be unreachable, got %s", exampleConnection)
+		}
+	})
+	t.Run("Connection bitmask handles Direct correctly", func(t *testing.T) {
+		exampleConnection := Connection(0)
+		exampleConnection.AddMethod(Direct)
+
+		if !exampleConnection.hasMethod(Direct) {
+			t.Errorf("expected Direct, got %s", exampleConnection)
+		}
+		exampleConnection.ClearMethod(Direct)
+		if exampleConnection != Unreachable {
+			t.Errorf("cleared connection should be unreachable, got %s", exampleConnection)
+		}
+	})
+
+	t.Run("Connection bitmask handles multiple flags correctly", func(t *testing.T) {
+		exampleConnection := Connection(0)
+		exampleConnection.AddMethod(Direct)
+		exampleConnection.AddMethod(Proxy)
+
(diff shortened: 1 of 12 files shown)


kubernetes/nodecollection.go
```diff
@@ -86,37 +87,94 @@ func (cns ClientsetNodeSource) NodeAddress(node *v1.Node) (string, int32, error)
...
 }
 
-// Endpoint an enumeration representing the various metrics endpoints
-type Endpoint uint8
+// Connection is a bitmask that describes the manner(s) in which
+// the agent can connect to an endpoint
+type Connection uint8
+
+const (
+	Unreachable Connection = 0
```
mnorbury, line 95: Good use of the default value. This might be better at the end of the `const` block as this will make the first mask value `1` rather than `2`.

kubernetes/nodecollection.go
```diff
@@ -86,37 +87,94 @@ func (cns ClientsetNodeSource) NodeAddress(node *v1.Node) (string, int32, error)
...
+	NodeStatsSummaryEndpoint Endpoint = "/stats/summary"
 
 	// NodeContainerEndpoint the /stats/container endpoint
-	NodeContainerEndpoint
+	NodeContainerEndpoint Endpoint = "/stats/container"
 
-	// NodeCadvisorEndpoint the /metrics/cadvisor/prometheus endpoint
-	NodeCadvisorEndpoint
+	// NodeCadvisorEndpoint the /metrics/cadvisor endpoint
+	NodeCadvisorEndpoint Endpoint = "/metrics/cadvisor"
```
mnorbury, line 137: I like using the Endpoint const to encode the endpoint strings 🏅

kubernetes/nodecollection.go
```diff
@@ -86,37 +87,94 @@ func (cns ClientsetNodeSource) NodeAddress(node *v1.Node) (string, int32, error)
...
-func (m EndpointMask) Available(endpoint Endpoint) bool {
-	_, ok := m[endpoint]
-	return ok
+// Available gets the availability of an endpoint for the specified connection method
+func (m EndpointMask) Available(endpoint Endpoint, method Connection) bool {
+	e, ok := m[endpoint]
+	if !ok {
+		return false
+	}
+	return e.HasMethod(method)
```
mnorbury, line 161: I think this can be the same as the other methods i.e. don't check the `ok` flag, just make use of the Connection default value if it isn't set.

kubernetes/nodecollection_test.go
```diff
@@ -89,46 +88,211 @@ func TestEnsureNodeSource(t *testing.T) {
 
 		ka, err := kubernetes.EnsureNodeSource(ka)
```
mnorbury, line 89: Not your code change, but this err should be checked before ka is used.

mnorbury (commented):
Nice work, I've left a few comments.

kubernetes/nodecollection.go
```diff
@@ -281,44 +318,85 @@ func (s sourceName) cadvisorMetrics() string {
...
 		prefix:   nd.prefix,
 		nodeName: nd.nodeName,
 	}
-	var err error
-
-	if mask.Available(NodeStatsSummaryEndpoint) {
-		// fetch stats/summary data
-		log.Debug("Fetching data from /stats/summary endpoint")
-		_, err = c.GetRawEndPoint(http.MethodGet, source.summary(), nd.workDir, api.statsSummary(), nil, true)
+	toFetch := map[Endpoint]bool{
```
rrobb6, line 328: Does this mean we're going to try and fetch all of the endpoints automatically every time?

rrobb6:
It looks good! I think encoding the connections at the level where we're going to use them makes more logical sense.

kubernetes/nodecollection.go
jpepin: Yes, this is consistent with the previous behavior as well

kubernetes/endpoint_test.go
```diff
@@ -0,0 +1,162 @@
...
+		exampleConnection.AddMethod(Direct)
+		exampleConnection.AddMethod(Proxy)
+
+		// should have both methods set
+		if !exampleConnection.hasMethod(Direct) {
+			t.Errorf("expected Direct, got %s", exampleConnection)
+		}
+		if !exampleConnection.hasMethod(Proxy) {
+			t.Errorf("expected Proxy, got %s", exampleConnection)
+		}
```
mnorbury, line 65: Maybe worth also showing in this sequence of tests that the following also works:

`!exampleConnection.hasMethod(Direct | Proxy)`

mnorbury (commented):
Looks good to me. I had another test case that I think is worth adding to demonstrate multiple flags can be checked at once with a bitmask.

mnorbury (approved):
Nice work! 🏅

PR #126, opened 2021-06-24
Update error logs

rrobb6:
#### What does this PR do?
Adds new error logs to the metrics-agent to clarify potential issues when installing the metrics-agent. 

#### Developer Done List

- [x] Tests Added/Updated
- [ ] Updated README.md

- [ ] Tests Pass Locally
- [ ] CI Build Passes

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -256,6 +256,7 @@ func (c httpMetricClient) retryWithBackoff(
 		var uploadURL, hash string
 		uploadURL, hash, err = c.GetUploadURL(metricFile, metricSampleURL, agentVersion, UID)
 		if err != nil {
+			log.Debugf("Client proxy or deployment YAML may be misconfigured.  Please check your client settings.")
 			log.Errorf("error encountered while retrieving upload location: %v", err)
 			continue
 		}
--- a/kubernetes/kubernetes.go
+++ b/kubernetes/kubernetes.go
@@ -9,8 +9,17 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/cloudability/metrics-agent/client"
+	"github.com/cloudability/metrics-agent/measurement"
+	k8s_stats "github.com/cloudability/metrics-agent/retrieval/k8s"
+	"github.com/cloudability/metrics-agent/util"
+	cldyVersion "github.com/cloudability/metrics-agent/version"
+	log "github.com/sirupsen/logrus"
 	"io"
 	"io/ioutil"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/clientcmd"
 	"net/http"
 	"net/url"
 	"os"
@@ -21,19 +30,10 @@ import (
 	"strings"
 	"time"
 
-	"github.com/cloudability/metrics-agent/client"
-	"github.com/cloudability/metrics-agent/measurement"
-	k8s_stats "github.com/cloudability/metrics-agent/retrieval/k8s"
 	"github.com/cloudability/metrics-agent/retrieval/raw"
-	"github.com/cloudability/metrics-agent/util"
-	cldyVersion "github.com/cloudability/metrics-agent/version"
-	log "github.com/sirupsen/logrus"
-	v1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/version"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
-	"k8s.io/client-go/tools/clientcmd"
 )
 
 //ClusterVersion contains a concatenated version number as well as the k8s version discovery info
@@ -90,8 +90,25 @@ const proxy = "proxy"
 const direct = "direct"
 const unreachable = "unreachable"
 
-//nolint llll
-const kbURL string = "https://support.cloudability.com/hc/en-us/articles/360008368193-Kubernetes-Metrics-Agent-Error-Messages"
+const kbTroubleShootingURL string = "https://help.apptio.com/en-us/cloudability/product/k8s-metrics-agent.htm"
+const kbProvisionURL string = "https://help.apptio.com/en-us/cloudability/product/k8s-cluster-provisioning.htm"
+const apiEndpoint string = "https://metrics-collector.cloudability.com"
+
+const forbiddenError string = uploadURIError + ": 403"
+const uploadURIError string = "Error retrieving upload URI"
+
+//nolint lll
+const transportError string = `Network transport issues are potentially blocking the agent from contacting the metrics collection API.
+	Please confirm that the metrics-agent is able to establish a connection to: %s`
+//nolint lll
+const apiKeyError string = `Current Cloudability API Key is expired and access needs to be re-enabled before re-provisioning the metrics-agent as detailed here: %s.
+	Please contact support to re-activate the API keys.
+	Note: Be sure to use the exact same cluster name as what is currently in use.
+	***IMPORTANT*** If the cluster is managed by GKE - there are special instructions for provisioning.`
+const rbacError string = `RBAC role in the Cloudability namespace may need to be updated.
+	Re-provision the metrics-agent for this cluster as detailed here: ` + kbProvisionURL +
+	`Note: Be sure to use the exact same cluster name as what is currently in use.
+	***IMPORTANT*** If the cluster is managed by GKE - there are special instructions for provisioning.`
 
 //CollectKubeMetrics Collects metrics from Kubernetes on a predetermined interval
 func CollectKubeMetrics(config KubeAgentConfig) {
@@ -120,8 +137,9 @@ func CollectKubeMetrics(config KubeAgentConfig) {
 	err := fetchDiagnostics(kubeAgent.Clientset, config.Namespace, kubeAgent.msExportDirectory)
 
 	if err != nil {
+		log.Warnf(rbacError)
 		log.Warnf("Warning non-fatal error: Agent error occurred retrieving runtime diagnostics: %s ", err)
-		log.Warnf("For more information see: %v", kbURL)
+		log.Warnf("For more information see: %v", kbTroubleShootingURL)
 	}
 
 	err = downloadBaselineMetricExport(kubeAgent, clientSetNodeSource)
@@ -358,8 +376,20 @@ func (ka KubeAgentConfig) sendMetrics(metricSample *os.File) {
 
 	err = SendData(metricSample, ka.clusterUID, cldyMetricClient)
 	if err != nil {
+		if warnErr := handleError(err); warnErr != "" {
+			log.Warnf(warnErr)
+		}
 		log.Fatalf("error sending metrics: %v", err)
 	}
+}
+
+func handleError(err error) string {
+	if err.Error() == forbiddenError {
+		return fmt.Sprintf(apiKeyError, kbProvisionURL)
+	} else if strings.Contains(err.Error(), uploadURIError) {
+		return fmt.Sprintf(transportError, apiEndpoint)
+	}
+	return ""
 }
 
 // SendData takes Cloudability metric sample and sends data to Cloudability via go client
@@ -587,8 +617,7 @@ func ensureMetricServicesAvailable(config KubeAgentConfig) (KubeAgentConfig, err
 	if config.RetrieveNodeSummaries {
 		config, err = ensureNodeSource(config)
 		if err != nil {
-			log.Warnf("Warning non-fatal error: Agent error occurred verifying node source metrics: %s ", err)
-			log.Warnf("For more information see: %v", kbURL)
+			log.Warnf(handleNodeSourceError(err))
 		} else {
 			log.Infof("Node summaries connection method: %s", config.NodeMetrics.Options(NodeStatsSummaryEndpoint))
 			log.Infof("Node container metrics connection method: %s", config.NodeMetrics.Options(NodeContainerEndpoint))
@@ -608,10 +637,24 @@ func ensureMetricServicesAvailable(config KubeAgentConfig) (KubeAgentConfig, err
 	}
 
 	if !config.RetrieveNodeSummaries && !config.CollectHeapsterExport {
+		//nolint lll
+		log.Debugf(`Unable to retrieve data due to metrics-agent configuration.
+			May be caused by cluster security mis-configurations or the RBAC role in the Cloudability namespace needs to be updated.
+			Please confirm with your cluster security administrators that the RBAC role is able to work within your cluster's security configurations.`)
 		return config, fmt.Errorf("unable to retrieve node summaries or heapster export: %s", err)
 	}
 
 	return config, nil
+}
+
+func handleNodeSourceError(err error) string {
+	var nodeError string
+	if strings.Contains(err.Error(), "Please verify RBAC roles") {
+		nodeError = rbacError
+	}
+	errStr := "Warning non-fatal error: Agent error occurred verifying node source metrics: %v\n" +
+		"For more information see: %s"
+	return nodeError + fmt.Sprintf(errStr, err, kbTroubleShootingURL)
 }
 
 func createKubeHTTPClient(config KubeAgentConfig) (KubeAgentConfig, error) {
--- a/kubernetes/kubernetes_unit_test.go
+++ b/kubernetes/kubernetes_unit_test.go
@@ -1,0 +1,38 @@
+package kubernetes
+
+import (
+	"fmt"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"testing"
+)
+
+func TestKubernetes(t *testing.T) {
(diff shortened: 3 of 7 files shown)


glovvorn-app (dismissed):
LGTM

client/client.go
```diff
@@ -256,6 +256,7 @@ func (c httpMetricClient) retryWithBackoff(
 		var uploadURL, hash string
 		uploadURL, hash, err = c.GetUploadURL(metricFile, metricSampleURL, agentVersion, UID)
 		if err != nil {
+			log.Debugf("Client proxy or deployment YAML may be misconfigured.  Please check your client settings.")
```
boharra, line 259: Will we get Debug logs from the clients? Is that level enabled by default on their deployments or is that something we'd have to ask them to change to be able to get this additional info?

kubernetes/kubernetes.go
```diff
@@ -358,10 +377,20 @@ func (ka KubeAgentConfig) sendMetrics(metricSample *os.File) {
 
 	err = SendData(metricSample, ka.clusterUID, cldyMetricClient)
 	if err != nil {
+		log.Warnf(handleError(err))
```
boharra, line 380: I think we only want to log this is handleError() return is not empty

kubernetes/kubernetes.go
```diff
@@ -608,12 +636,27 @@ func ensureMetricServicesAvailable(config KubeAgentConfig) (KubeAgentConfig, err
...
+			"RBAC role is able to work within your cluster's security configurations.")
 		return config, fmt.Errorf("unable to retrieve node summaries or heapster export: %s", err)
 	}
 
 	return config, nil
 }
 
+func handleNodeSourceError(err error) string {
+	var error string
+	if strings.Contains(err.Error(), "Please verify RBAC roles") {
```
boharra, line 652: I like these options to add additional info to the logs where we know more about the problem 👍

kubernetes/kubernetes_unit_test.go
```diff
@@ -0,0 +1,38 @@
...
+			errorStr := handleError(fmt.Errorf("test error"))
+			Expect(errorStr).To(Equal(""))
+		})
+
+		It("Node source error handler should return an error if we need to verify RBAC roles", func() {
+			errorStr := handleNodeSourceError(fmt.Errorf("Please verify RBAC roles"))
+			Expect(errorStr).To(ContainSubstring("RBAC role in the Cloudability namespace may need to be updated."))
+		})
+	})
+})
```
cadams8238, line 38: Yay, unit tests! 🙌

kubernetes/kubernetes.go
```diff
@@ -91,7 +91,25 @@ const direct = "direct"
 const unreachable = "unreachable"
 
 //nolint llll
```
cadams8238, line 93: This is no longer doing anything useful. Would you mind removing please?

kubernetes/kubernetes.go
```diff
@@ -91,7 +91,25 @@ const direct = "direct"
...
+const apiKeyError string = "Current Cloudability API Key needs to be re-enabled before re-provisioning " +
+	"the metrics-agent as detailed here: %s.\n" +
+	"Please contact support to re-activate the API keys.\n" +
+	"Note 1: Be sure to use the exact same cluster name as what is currently in use.\n" +
+	"Note 2: If the cluster is managed by GKE - there are special instructions for provisioning"
+const rbacError string = "RBAC role in the Cloudability namespace may need to be updated.\n" +
+	"Re-provision the metrics-agent for this cluster as detailed here: %s\n" +
+	"Note: Be sure to use the exact same cluster name as what is currently in use.\n" +
+	"***NOTE*** Please note that if the cluster is managed by GKE, " +
+	"there are special instructions for provisioning.\n"
```
cadams8238, line 112: ⭐ 
I think these error messages would be slightly easier to read if we used multiline strings instead of concatenation. What do you think?

Also, can we be consistent with how we're adding notes in these errors? There are 3 different formats currently - `Note 1/2`, `Note`, and `***NOTE***`...Hmm looking closer, the `apiKeyError` and `rbacError` both have the same notes about GKE and same cluster naming - what do you think about cleaning up both of these error messages and use the same messaging for both error's "notes"? I'm thinking this "notes" string (kinda like an error "footer" to be suffixed on each error) could look something like: 
```
Be sure to use the exact same cluster name as what is currently in use. 
***Please note that if the cluster is managed by GKE, there are special instructions for provisioning.***
```

kubernetes/nodecollection.go
```diff
@@ -110,7 +110,8 @@ func downloadNodeData(prefix string,
 
 	for _, n := range nodes {
 		if n.Spec.ProviderID == "" {
-			failedNodeList[n.Name] = errors.New("Provider ID for node does not exist. " +
+			log.Warnf("Node ProviderID is not set, may cause inconsistent gathering of metrics data.")
+			failedNodeList[n.Name] = errors.New("provider ID for node does not exist. " +
```
boharra, line 114: Do we miss these failedNodeList errors in the logs? The new log statement seems similar, and I'm just wondering if we'll be double logging the same issue here when Nodes fail (because I think parent calls to this method log the error). Is there some more nuance to this to consider when we'd be trying to triage a customer issue where the additional log statement tells us more?

kubernetes/kubernetes.go
```diff
@@ -120,8 +138,9 @@ func CollectKubeMetrics(config KubeAgentConfig) {
 	err := fetchDiagnostics(kubeAgent.Clientset, config.Namespace, kubeAgent.msExportDirectory)
 
 	if err != nil {
+		log.Warnf(rbacError, kbProvisionURL)
```
cadams8238, line 141: ⭐ 
I think this could be clearer if `rbacError` was updated to include the provisioning url directly. My thinking is, the `rbacError` will always need to include the link to docs on provisioning which gets rid of the need to add functionality to support another string with the `%s` (in the `rbacError` string). We could always add that support in later if we need it. I'd imagine the error message would need to be tweaked if that's the case.

To illustrate what I'm thinking:
```
const rbacError string = `RBAC role in the Cloudability namespace may need to be updated.
	Re-provision the metrics-agent for this cluster as detailed here: ` + kbProvisionURL + [rest of string]
```

kubernetes/kubernetes.go
```diff
@@ -120,8 +138,9 @@ func CollectKubeMetrics(config KubeAgentConfig) {
...
 		log.Warnf("Warning non-fatal error: Agent error occurred retrieving runtime diagnostics: %s ", err)
-		log.Warnf("For more information see: %v", kbURL)
+		log.Warnf("For more information see: %v", kbErrorURL)
```
cadams8238, line 143: ⭐ 
The naming for this var:`kbErrorURL` can cause confusion. Originally I thought this meant it was the URL to the error occurring but after looking at the link, it just goes to our help docs. What do you think about updating the naming to something like `kbTroubleshootDocsURL` or `kbHelpDocsURL`.

kubernetes/kubernetes.go
```diff
@@ -608,12 +636,27 @@ func ensureMetricServicesAvailable(config KubeAgentConfig) (KubeAgentConfig, err
...
 }
 
+func handleNodeSourceError(err error) string {
+	var error string
+	if strings.Contains(err.Error(), "Please verify RBAC roles") {
+		error = fmt.Sprintf(rbacError, kbProvisionURL)
+	}
+	errStr := "Warning non-fatal error: Agent error occurred verifying node source metrics: %v\n" +
+		"For more information see: %s"
+	return error + fmt.Sprintf(errStr, err, kbErrorURL)
```
cadams8238, line 657: ⭐ 
Ah, nice. I like the way you coded this so if the error is not related to RBAC roles, then the final error will be prefixed by an empty string - clever! 😄 One suggestion for improvement: It's a bit confusing to read though with the `error` var naming as-is. What do you think about changing that var to be something more descriptive, like: `optionalRBACErrMsg` or `optionalProvisioningErrMsg`

kubernetes/kubernetes.go
```diff
@@ -608,12 +636,27 @@ func ensureMetricServicesAvailable(config KubeAgentConfig) (KubeAgentConfig, err
 	}
 
 	if !config.RetrieveNodeSummaries && !config.CollectHeapsterExport {
+		log.Debugf("Unable to retrieve data due to metrics-agent configuration.\n" +
+			"May be caused by cluster security mis-configurations or the RBAC role in the Cloudability " +
+			"namespace needs to be updated.\n" +
+			"Please confirm with your cluster security administrators that the " +
+			"RBAC role is able to work within your cluster's security configurations.")
```
cadams8238, line 643: Similar to the error strings above, what do you think about changing these strings to use a back- ticked, multiline string?

kubernetes/kubernetes.go
cadams8238, line 652: Do we care about case sensitivity when checking if the error contains this string?

kubernetes/nodecollection.go
```diff
@@ -429,7 +430,12 @@ func allowDirectConnect(config KubeAgentConfig, nodes []v1.Node) bool {
 	// if any Fargate nodes are found.
 	for _, n := range nodes {
 		if isFargateNode(n) {
-			log.Infof("Fargate node found in cluster, direct node connection disabled. Learn more about Fargate support: %s", kbURL) //nolint: lll
+			log.Infof("Direct connection to an AWS Fargate node is not possible, so when a Fargate node is " +
+				"detected in the cluster, the agent switches to using kube-proxy for all node connections " +
+				"in the cluster. The allocation of EKS Fargate tasks are not currently supported via the " +
+				"containers feature dashboard. " +
+				"Please contact your technical account manager for instructions on how to access" +
+				"EKS Fargate information via the reporting feature.") //nolint: lll
```
cadams8238, line 438: Similar to above, what do you think about changing these to a back-ticked, multiline string? (And I think that would get rid of the `nolint` too)

kubernetes/nodecollection_test.go
```diff
@@ -541,7 +541,7 @@ func TestDownloadNodeData(t *testing.T) {
 			t.Error("Expected error for nodename \"proxyNode\"")
 		}
 
-		if errFromList.Error() != "Provider ID for node does not exist. "+
+		if errFromList.Error() != "provider ID for node does not exist. "+
```
cadams8238, line 544: I see this string is used twice (once in the code/once in this test) - do you think it's worth creating a variable?

kubernetes/kubernetes.go
```diff
@@ -91,7 +91,25 @@ const direct = "direct"
 const unreachable = "unreachable"
 
 //nolint llll
-const kbURL string = "https://support.cloudability.com/hc/en-us/articles/360008368193-Kubernetes-Metrics-Agent-Error-Messages"
+const kbErrorURL string = "https://help.apptio.com/en-us/cloudability/product/k8s-metrics-agent.htm"
+const kbProvisionURL string = "https://help.apptio.com/en-us/cloudability/product/k8s-cluster-provisioning.htm"
+const apiEndpoint string = "https://metrics-collector.cloudability.com"
+
+const forbiddenError string = "Error retrieving upload URI: 403"
```
cadams8238, line 98: This could be DRYed up some using the uploadURIError string like so:
```
const forbiddenError string = uploadURIError + ": 403"
```

client/client.go
boharra: I see it's done elsewhere in the code, so maybe it is on by default...nbd if you're not sure, I'm sure it's fine. At this point I'm just curious if you already know how we handle debug level logging in the metrics agent.

rrobb6: They'd have to enable it, but they can do it in the deployment and then let the pod restart and see this error.  The Readme mentions it as one of the configuration options.

boharra:
Looks good! Left some minor comments/questions, but overall I think these will help us a lot and I like that you were able to test your error handling!

cadams8238 (changes requested):
Nice work on this! I riddled the code with a bunch of comments, only some are blocking to approval though. I'll leave a ⭐ on the ones I think are the most important - the rest are stylistic and nit-picky so feel free to ignore or do what you'd like with em.

kubernetes/nodecollection.go
rrobb6: The second error will output a list of all of the nodes that have errors with the nodes.  I'll update this message to make it clearer because this usually happens in cases where the node is in a self managed environment (i.e. not on AKS, EKS, or GKE)

kubernetes/kubernetes.go
rrobb6: Yeah it goes to our Kubernetes documents where our typical errors are provided for customer viewing.  That's why I named it `kbErrorURL` because we know that's where the public error information is rather than the provisioning information.

kubernetes/kubernetes.go
rrobb6: I think the `***NOTE***` is to indicate that this is a really important thing to pay attention to.  I can make the other one consistent though.

kubernetes/kubernetes.go
rrobb6: Yes, it should match that case.

kubernetes/nodecollection.go
boharra: Sounds good, adding a little color to the log.Warn message will make it more useful to read in addition to the list of failed nodes with the more generic "provider id doesn't exist, may cause more errors"

kubernetes/kubernetes.go
rrobb6: I'll try, but a lot of these methods have really high cyclometric complexity as it is, so if I add another if statement, it might irritate the linter.  I don't think that will happen in this case, but might need to refactor.

kubernetes/kubernetes.go
rrobb6: Also, a lot of the concatenation is because they're one long sentence, so the linter gets cranky because of the length.  I'll see if I can make them multiline (and get rid of the newlines), but it doesn't make sense to break full up sentences into multiple lines.

kubernetes/kubernetes.go
rrobb6: Ah right, that's a reserved word, I'll fix it.

kubernetes/kubernetes.go
```diff
@@ -91,7 +91,25 @@ const direct = "direct"
...
+const uploadURIError string = "Error retrieving upload URI"
+const transportError string = "Network transport issues are potentially blocking the agent from " +
+	"contacting the metrics collection API.\n" +
+	"Please confirm that the metrics-agent is able to establish a connection to: %s"
+const apiKeyError string = "Current Cloudability API Key needs to be re-enabled before re-provisioning " +
```
bbylovas, line 103: Would change "Current Cloudability API Key needs to be re-enabled before reprovisioning the metrics-agent as detailed here: %s/n" to something like "Current Cloudability API Key is expired.  For more details see %s.\n". It's a slight correction since we aren't reactivating the same api key, but generating a new one

kubernetes/nodecollection_test.go
rrobb6: I think with just two uses and being so short, it's fine as is.

kubernetes/kubernetes.go
rrobb6: I think the provisioning page doesn't have info about expired API keys, so we need some info about re-activating the expired key, because part of these messages is that, customers might not actually go to the help pages, so we want to give as much info in as short of a blurb as possible.

bbylovas: That's a good point. I think we should make sure that we're not saying we are reactivating the same api key, but instead generating a new one. We've gotten customer questions in the past about the particularities of provisioning so I just want to make sure our messaging is consistent about what is actually happening.

kubernetes/nodecollection.go
```diff
@@ -429,7 +432,12 @@ func allowDirectConnect(config KubeAgentConfig, nodes []v1.Node) bool {
 	// if any Fargate nodes are found.
 	for _, n := range nodes {
 		if isFargateNode(n) {
-			log.Infof("Fargate node found in cluster, direct node connection disabled. Learn more about Fargate support: %s", kbURL) //nolint: lll
+			log.Infof("Direct connection to an AWS Fargate node is not possible, so when a Fargate node is " +
+				"detected in the cluster, the agent switches to using kube-proxy for all node connections " +
```
bbylovas: I don't think our Proxy connection mode and using kube-proxy are the same thing. I believe when we are in the Proxy mode, we are using kube-api-server as a proxy to collect metrics. Would change "...the agent switches to using kube-proxy for..." to "...the agent's connection mode switches to proxy for..."

bbylovas (commented):
This is awesome work. Kudos on taking on the barrage of feedback as well 👍

kubernetes/kubernetes.go
rrobb6: How does this sound?

> Current Cloudability API Key is expired and access needs to be re-enabled before re-provisioning the metrics-agent as detailed here

bbylovas: Love it 👍

kubernetes/nodecollection.go
rrobb6: Fixed

cadams8238 (approved):
LGTM!

PR #130, opened 2021-08-05
Tidy up go.mod

cadams8238:
#### What does this PR do?
This PR is for fixing the e2e tests. The tests were not running successfully after a change was recently made to GH actions that no longer supported the CNI plugin we used previously. The CNI plugin we used was an indirect dependency so ultimately we needed to upgrade the k8s package dependencies to their latest versions - v0.22.1 - which fixed the issue. These upgrades warranted some code changes, namely:
* The CoreV1 API changed some function signatures to require a context being passed in. The majority of the changes you see in this PR are due to this update. **No functionality was changed by adding a context.** In future PRs, we can choose to do more with the context, but for now it's acting more as an initial placeholder to build off of later.
* The stats API changed which package the type Summary lives - from `k8s.io/kubernetes/pkg/kubelet/apis/stats/v1alpha1` => `k8s.io/kubelet/pkg/apis/stats/v1alpha1`

Additionally, I pinned the e2e test dependencies in `tools.go` so that `go mod tidy` can be run successfully without removing necessary dependencies, making development smoother. Tidying would remove those dependencies because they were not top-level dependencies.

One more tiny change was for the go linter. Golangci-lint no longer uses `golint` as it's default linter, but instead uses `revive`. Those changes are reflected in this PR.

go.mod

N/A

Quite a few modules were outdated/unused and were throwing security vulnerabilities. This is the first effort at remediating those vulnerabilities. A Snyk scan will be run after this is merged to see how many it fixes.

cleaning

N/A

#### Developer Done List

- [ ] Tests Added/Updated
- [ ] Updated README.md
- [X] Verified backward compatible
- [X] Considered Security, Availability and Confidentiality

- [ ] Tests Pass Locally
- [ ] CI Build Passes


go.mod
```diff
@@ -3,73 +3,19 @@ module github.com/cloudability/metrics-agent
...
-	k8s.io/component-base v0.0.0 => k8s.io/component-base v0.17.3
-	k8s.io/cri-api v0.0.0 => k8s.io/cri-api v0.17.3
-	k8s.io/csi-translation-lib v0.0.0 => k8s.io/csi-translation-lib v0.17.3
-	k8s.io/kube-aggregator v0.0.0 => k8s.io/kube-aggregator v0.17.3
-	k8s.io/kube-controller-manager v0.0.0 => k8s.io/kube-controller-manager v0.17.3
-	k8s.io/kube-proxy v0.0.0 => k8s.io/kube-proxy v0.17.3
-	k8s.io/kube-scheduler v0.0.0 => k8s.io/kube-scheduler v0.17.3
-	k8s.io/kubectl v0.0.0 => k8s.io/kubectl v0.17.3
-	k8s.io/kubelet v0.0.0 => k8s.io/kubelet v0.17.3
-	k8s.io/legacy-cloud-providers v0.0.0 => k8s.io/legacy-cloud-providers v0.17.3
```
boharra, line 72: Nice to not have this giant k8s replace block, seems like this cleanup was a long time coming

kubernetes/kubernetes.go
```diff
@@ -236,8 +241,8 @@ func newKubeAgent(config KubeAgentConfig) KubeAgentConfig {
 	return config
 }
 
-func (ka KubeAgentConfig) collectMetrics(
-	config KubeAgentConfig, clientset kubernetes.Interface, nodeSource NodeSource) (rerr error) {
+func (ka KubeAgentConfig) collectMetrics(ctx context.Context, config KubeAgentConfig,
```
rrobb6, line 244: Looks like this method has developed a code smell (needs to be refactored to reduce complexity)

kubernetes/kubernetes.go
```diff
@@ -30,6 +22,16 @@ import (
 	"strings"
 	"time"
 
+	"github.com/cloudability/metrics-agent/client"
```
rrobb6, line 25: Looks like our imports are batching together weird.  Should we make it consistent so that they're listed together and alphabetical?

boharra (dismissed):
Looks good to me, I'm not very familiar with metrics agent code but this seems like a safe upgrade! 🚢

kubernetes/kubernetes.go
```diff
@@ -172,12 +177,13 @@ func CollectKubeMetrics(config KubeAgentConfig) {
...
 
 		case <-pollChan.C:
-			err := kubeAgent.collectMetrics(kubeAgent, kubeAgent.Clientset, clientSetNodeSource)
+			err := kubeAgent.collectMetrics(ctx, kubeAgent, kubeAgent.Clientset, clientSetNodeSource)
 			if err != nil {
 				log.Fatalf("Error retrieving metrics %v", err)
 			}
 
 		case <-doneChan:
+			ctx.Done()
```
rrobb6, line 186: Thanks for making sure this gets cleaned up!

testdata/e2e/e2e_helper.go
```diff
@@ -16,7 +16,7 @@ import (
 	"github.com/prometheus/prom2json"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	statsapi "k8s.io/kubernetes/pkg/kubelet/apis/stats/v1alpha1"
+	statsapi "k8s.io/kubelet/pkg/apis/stats/v1alpha1"
```
rrobb6, line 19: These are equivalent packages?  Did it just get moved?

rrobb6 (commented):
Pretty straightforward changes.  Noticed a couple of code smells being reported by sonarcloud.

go.mod
cadams8238: Seriously. It's much cleaner now ✨

testdata/e2e/e2e_helper.go
cadams8238: Yup, they're equivalent for our purposes. We just use the `Summary` type which is the same in both. From what I was reading, they wanted to move sub packages out of the main kubernetes package - I can't remember what their reasoning was off the top of my head.

kubernetes/kubernetes.go
cadams8238: Yeah VSCode is awesome in this - I have a go format tool installed that utilizes `goimports`. It does all the work that `gofmt` does, like making sure your code follows the Go standard, as well as removing unused imports and  organizing imports alphabetically with standard libs grouped at the top. (If you put newlines in-between imports then it'll organize each chunk alphabetically.) 

Do y'all not use that tool in GoLand? I can go around to each file and do clean up to make them all consistent if that'd be helpful. Shouldn't take too long.

kubernetes/kubernetes.go
cadams8238: Thanks for catching! I'll take a look and update.

kubernetes/kubernetes.go
rrobb6: I have that, but I was thinking about the newline between groups - is that a standard we want to keep?

cadams8238: Oh I see. I misunderstood your question. I have no idea why there's a newline. Happy to remove.

boharra: I kinda like the newline between groups, shows which imports are custom vs go standard go libs. But up to y'all!

cadams8238: I agree. I like the newline between the standard libs and third party packages. I think Becky means the additional newline below separating third party packages iiuc.

rrobb6: I was just wondering about newlines in general.  But at the very least, we probably should only have at least one for between standard and custom packages.

kubernetes/kubernetes.go
cadams8238: Hmm looking at those code smells Sonar threw, fixing them doesn't seem within the scope of this PR. Looks like they weren't caught with the initial writing of that function (or maybe if/when someone added to it?). To fix each would divert time and resources away from the 1.21 work. I agree with you - ideally we'd fix and not have large functions like this in the codebase. Given it's lower priority than getting the 1.21 work out the door, I'm going to punt on this for now. If there's time later, we can revisit.

Code smells:
Screen Shot 2021-09-09 at 11 03 43 AM
Screen Shot 2021-09-09 at 11 03 52 AM

kubernetes/kubernetes.go
cadams8238: Sounds good! I'll remove and push up those changes.

rrobb6 (approved):
LGTM :shipit:

PR #151, opened 2021-12-16
Remove temp local cache in Dockerfile to resolve x509 certificate error

cadams8238:
#### What does this PR do?
It removes the temp local cache directory in the Dockerfile to resolve a x509 certificate signed by unknown authority error encountered by a customer. The plan is to cut a beta release and have the customer test the fix on their cluster (since we can not reproduce the issue locally). If it's successful, we'll officially release the change to GA.

Dockerfile

By spinning up the Metrics Agent like usually and seeing if it collects metrics successfully.

Like mentioned above, a customer was running into the x509 error which was the catalyst for this fix. I researched the error and it was resounding across all different sources that adding the remove cmd in the Dockerfile solved the error. 
[Source1](https://github.com/aws/aws-sdk-go/issues/2322#issuecomment-443502850)
[Source2](http://pinter.org/archives/8173)
[Source3](https://hackernoon.com/alpine-docker-image-with-secured-communication-ssl-tls-go-restful-api-128eb6b54f1f)
[Source4](https://stackoverflow.com/questions/52601404/aws-ses-error-x509-certificate-signed-by-unknown-authority) - Altho slightly different. They're using a version of Alpine that's >=3.3 so they're able to use the `no-cache` flag. That flag is the equilvalent of running the update cmd + removing the cache per this source [here](https://github.com/gliderlabs/docker-alpine/blob/master/docs/usage.md#disabling-cache).

And note that in our Dockerfile, we use the `update` flag instead of `update-cache` but they are essentially the same thing. See [here](https://stackoverflow.com/questions/43681432/explanation-of-the-update-add-command-for-alpine-linux) for more details.

N/A

#### Developer Done List

- [ ] Tests Added/Updated
- [ ] Updated README.md
- [X] Considered Security, Availability and Confidentiality

- [ ] Tests Pass Locally
- [x] CI Build Passes
- [x] Verified changes are backward compatible

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -12,6 +12,7 @@
 
 #### Developer Done List
 
+- [ ] Incremented verison.go
 - [ ] Tests Added/Updated
 - [ ] Updated README.md
 - [ ] Verified backward compatible
--- a/deploy/docker/Dockerfile
+++ b/deploy/docker/Dockerfile
@@ -19,7 +19,7 @@ ARG application
 
 
 # Allow delve to run on Alpine based containers.
-RUN apk --update upgrade && apk add ca-certificates
+RUN apk --update upgrade && apk add ca-certificates && rm -rf /var/cache/apk/*
 
 RUN addgroup -g 1000 agent && \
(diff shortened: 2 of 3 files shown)


bbylovas (approved):
LGTM, nice digging 👍

PR #183, opened 2022-06-16
Add flag to parse/sanitize data from core files

jdhudson3:
#### What does this PR do?
Adds a flag that will parse and remove much non-essential data from the metrics agent. This decreases core file sizes by approximately 15%, but may differ depending on customer cluster configurations.

The performance impact of this on large clusters is currently undetermined, and enabling this flag carries risk of metrics agent slowdowns/crashes. If enabled extra headroom should be given to the agent to determine an appropriate cpu/memory footprint. This is currently flagged off by default

This can be observed to some extent via the added benchmark tests:

```
goos: darwin
goarch: amd64
pkg: github.com/cloudability/metrics-agent/retrieval/raw
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
BenchmarkMetricFileCreation
BenchmarkMetricFileCreation-16    	    1720	    724394 ns/op
BenchmarkPodsFile
BenchmarkPodsFile-16              	    1250	    864695 ns/op
BenchmarkParsedPodsFile
BenchmarkParsedPodsFile-16        	     573	   2097312 ns/op
PASS
```

Tested by adding agent to a minimal cluster, ensured that metrics were processed and no errors were observed.

Per customer request, this will decrease the amount of data gathered, including sanitization of environment variables.

#### Developer Done List

- [x] Tests Added/Updated
- [x] Updated README.md

- [ ] Tests Pass Locally
- [ ] CI Build Passes


retrieval/raw/raw_endpoint.go
```diff
@@ -114,13 +123,121 @@ func downloadToFile(c *Client, method, sourceName string, workDir *os.File, URL
...
+	for i, item := range cast.Items {
+
+		// stripping env var and related data from the object
+		item.ObjectMeta.ManagedFields = nil
+		if _, ok := item.ObjectMeta.Annotations[KubernetesLastAppliedConfig]; ok {
+			delete(item.ObjectMeta.Annotations, KubernetesLastAppliedConfig)
+		}
+		for j, container := range item.Spec.Containers {
+			container.Env = nil
+			item.Spec.Containers[j] = container
```
apptiopzuo, line 220: Specifically, the task should not only filter env variable. It should only keep the fields that Containers pipeline need. So, it should be something like:
sanitizedContainer.labels = container.labels
sanitizedContainer.name = container.name
etc...
The task includes also identifying the required fields to be kept.

apptiopzuo, line 220: It shouldn't only filter "env" variable. Instead, it should only extract/keep fields that needs by Containers Insight.
The task also includes identifying the fields need be kept.
For Pods, following fields should be kept in my opinion, but it needs be verified with Denormizer:
* name
* namespace
* node
* labels
* status
containers[]:
* id
* state
* limits
* requests

jdhudson3: Hey Peter. The method of serialization/deserialization used here does remove a decent amount of non-relevant data in addition to the env specific filtering done above by using the structs we also use internally for parsing this data. This ticket accomplishes the primary task as requested by the customer as well as provides the ability to easily decrease our data collected in the future.

While it would be nice to do a full audit of the data used by our internal processes and pare down to 
*exactly* what we need, I do not believe that is a reasonable goal at this moment. I would be happy to discuss the specific challenges of this internally if you'd like.

retrieval/raw/raw_endpoint.go
```diff
@@ -114,13 +124,145 @@ func downloadToFile(c *Client, method, sourceName string, workDir *os.File, URL
...
+	container.Args = nil
+	container.ImagePullPolicy = ""
+	container.LivenessProbe = nil
+	container.StartupProbe = nil
+	container.ReadinessProbe = nil
+	container.TerminationMessagePath = ""
+	container.TerminationMessagePolicy = ""
+	container.SecurityContext = nil
+	return container
+}
```
jyin-apptio, line 253: It makes me concern that we start a risk pattern to manually set a lot fields to nil or "". TBH, I didn't verify whether the logic is correct or not unless I cross-check denor logic which is time consuming. If you use any methodology to get these fields, different logic in different files, etc, I am glad to hear.
Here some risks in my head.
1. It is low readability, for example, the reader will always ask why do we need to set  ManagedFields to nil. (I guess all these field are not used by our backend, but that is not clear from the code)
2. It is risk if k8s data type is updated and how do we guarantee the backward compatible. For example, if one future version removes/rename the field container.SecurityContext etc
3. It is difficult to get what data we need/collect.. We could only tell what data we don't need and change to empty.  It also indicates this logic is difficult to test.

I remember we discuss define our own model (you could still refer some k8s type), and only list the fields we need or potential need, so that we could directly UnMarshal raw data to that model we need ( don't need to manually hack abc field to be empty), it will have better performance. Also, that will be self explain what data we collect from customer(easy to communicate to new client), easy to maintain(if we need to collect more data/field, just add new field in that model, is there any challenge blocking you from that option

retrieval/raw/raw_endpoint.go
```diff
@@ -114,13 +124,145 @@ func downloadToFile(c *Client, method, sourceName string, workDir *os.File, URL
...
+		return err
+	}
+	to = sanitizeData(out.Elem().Interface())
+
+	data, err := json.Marshal(to)
+	if err != nil {
+		return err
+	}
+	_, err = io.Copy(writer, bytes.NewReader(data))
+	return err
```
jyin-apptio, line 158: Do you think we need add some error information to help if err happens? Like unable to parse which file, not supported type and whatever.

retrieval/raw/raw_endpoint.go
jdhudson3: The initial approach only sanitized potentially sensitive data that we very obviously did not need. Peter requested that I do a more in-depth pass to remove all data that we did not need, and this is the data that I determined we do not need. I could potentially create our own model for this, but that puts the burden on us to maintain it and ensure that it is in sync with what we're using on the backend (as we are using the kube object definitions there as well).

1. I agree, I could put a comment along the lines of "we don't use these", but I figured that was sort of implicit.
2. If it's removed and our code isn't updated, no changes/errors occur. If it's removed from the library when we update the version of the dependency, then it wont compile until the reference is removed.

I'll take a look at copying the k8s objects and seeing if that's something we want to maintain.

jdhudson3: So looking into this, it is definitely do-able, but I personally do not think it is something that we want to maintain.

We would effectively be recreating somewhere between 40-60% of the object definitions from [this file](https://github.com/kubernetes/kubernetes/blob/v1.24.2/pkg/apis/core/types.go).

In my experience, recreating complex contracts with 3rd parties becomes _very_ painful and time intensive as time goes on, and is not something I personally view as worth the hassle of maintaining. If this is something that the team thinks is worth doing, then I am open to pursuing it, but I would strongly advocate against it.

retrieval/raw/raw_endpoint.go
```diff
@@ -114,13 +124,145 @@ func downloadToFile(c *Client, method, sourceName string, workDir *os.File, URL
...
+		return sanitizeSelectorMatchedResourceList(to)
+	case PodList:
+		return sanitizePodList(to)
+	case LabelMapMatchedResourceList:
+		return sanitizeMapMatchedResourceList(to)
+	case NamespaceList:
+		return sanitizeNamespaceData(to)
+	}
+	return to
+}
```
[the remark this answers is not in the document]
jyin-apptio: Ok, get it. That is not obviously.. Why do we use two separate way to handle the files. A), only unmarshalling the data we need (defined in our struct). B) manually set some fields to empty after unmarshal

README.md
```diff
@@ -77,6 +78,7 @@ Flags:
       --force_kube_proxy                         When true, forces agent to use the proxy to connect to nodes rather than attempting a direct connection. Default: False
       --poll_interval int                        Time, in seconds, to poll the services infrastructure. Default: 180 (default 180)
       --namespace string                         The namespace which the agent runs in. Changing this is not recommended. (default `cloudability`)
+      --parse_metric_data bool                   When true, metrics will be parsed and env variables sanitized prior to upload. (default `false`)
```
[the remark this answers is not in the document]
jyin-apptio: I think my question is, the variable is used to define weather we parse the file in cluster.. It is not equivalent to whether we sanitizer env variable or not, so, the description is misleading in my view

retrieval/raw/raw_endpoint.go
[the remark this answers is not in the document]
jdhudson3: Ideally it would all be explicitly defined objects that live in this repo but as I mentioned below that is, in my opinion, and unmaintainable mountain of data/structs. These objects were copied from the denormalizer and define the baseline set of data that we will need further down the pipeline and removes a small amount of high level data. `spec` for example is a first level key that only contains 1 piece of data that we need (labelselectors) out of a large number of fields and is therefore easy to parse out.

The further steps taken below where certain fields are nil'd out were done to remove sensitive or unused data that is highly nested

README.md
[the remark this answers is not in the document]
jdhudson3: Ahhh gotcha, I'll update it.

retrieval/raw/raw_endpoint.go
jdhudson3: If we're blindly bumping major versions we have a more serious problem imo 😄

To your point though, a non-backwards-compatible change would accompany a new kubernetes release and we would notice that issue rather quickly in both the agent & denormalizer when testing it. At the moment we are using a relatively old version of the kube models internally.

I could also add a unit test to make sure we're within some percentage of the original file size.

jyin-apptio: Just want to better understand what's the `pain` you mentioned above. If k8s makes backward compatible change to the field we care, we don't need to make any change to our own model.. If it is non-backward compatible change, I assume we just need to change the json name in our own model(small effort). BUT, we have to change that anyway in our backend system anyway even we use k8s model when we extracting data and put in measure model

Assert the file size is good than nothing! I still hope one day we could reach a point that we assert the data we collect.

apptiopzuo: I preferred to the approach to explicitly keep the data we care than filtering out data we don't need. There could be in the future "extended" fields from those responses via k8s sdk. We may not notice the change, but data ingestion file size could be blown up.

README.md
```diff
@@ -32,25 +32,26 @@ Cloudability Metrics Agent currently does not support OpenShift, Rancher or On P
...
+| CLOUDABILITY_LOG_FORMAT              |                                                                               Optional: Format for log output (JSON,PLAIN) Default: PLAIN                                                                                |
+| CLOUDABILITY_LOG_LEVEL               |                                                                     Optional: Log level to run the agent at (INFO,WARN,DEBUG,TRACE). Default: `INFO`                                                                     |
+| CLOUDABILITY_SCRATCH_DIR             |            Optional: Temporary directory that metrics will be written to. If set, must assure that the directory exists and that the user agent UID 1000 has read/write access to the folder. Default: `/tmp`            |
+| CLOUDABILITY_PARSE_METRIC_DATA       |                 Optional: Determines if metric files should be parsed and sanitized of env variables prior to upload. If set, may cause a significant increase in resource consumption. Default: `false`                 |
```
jyin-apptio, line 54: Could you change this place and `[cmd/kubernetesCmd.go]` as well

jyin-apptio (approved):

jdhudson3:
Closed by https://github.com/cloudability/metrics-agent/pull/189

PR #194, opened 2022-08-18
Gather K8s Resources Using Informers

### LINKED ISSUE #191: Intermittent liveness probes due to short "timeoutSeconds" on probe
Hi there,

We are seeing intermittent issues recently due to containers restarting with liveness probe failures:

```Liveness probe failed:```

We have tracked this down to the timeoutSeconds, which defaults to 1 second in k8s 1.20+.

A workaround is to change the values file, can we suggest you update the default [values file](https://github.com/cloudability/metrics-agent/blob/5e94e9d4c2ac0122005262bd7ae819699da6f5e0/charts/metrics-agent/values.yaml#L35) to add

timeoutSeconds: 5

We think this is due to runc taking a second or two to run the command.

### PR DESCRIPTION
**Pull request by daniel-spray**
#### What does this PR do?
Changes the way the metrics-agent collects the lists of Kubernetes resources on the cluster using Informers. This reduces the amount of calls to the api-server
Which included modifications to continue to support the parse sensitive metrics optional flag as we no longer use raw_endpoint for k8s resources.
retrieval/k8s/k8s_stats.go
Tested changes on a test cluster as well as passes e2e tests
Nope
Closes #191 
#### Developer Done List

- [x] Tests Added/Updated
- [x] Updated README.md

- [ ] Tests Pass Locally
- [ ] CI Build Passes

### CODE REVIEW

File: retrieval/k8s/k8s_stats.go [hunk 1]
```diff
@@ -1,82 +1,102 @@
...
+}
 
-	// get v1 App sources
-	for _, v1Apps := range v1AppSources {
-		_, err := rawClient.GetRawEndPoint(http.MethodGet, v1Apps, workDir, clusterHostURL+"/apis/apps/v1/"+v1Apps, nil, true)
+//GetK8sMetricsFromInformer loops through all k8s resource informers in kubeAgentConfig writing each to the WSD
+func GetK8sMetricsFromInformer(informers map[string]*cache.SharedIndexInformer, workDir *os.File) error {
+	for resourceName, informer := range informers {
+		// Cronjob informer will be nil if k8s version is less than 1.21, if so skip getting the list of cronjobs
+		if *informer == nil {
```
jdhudson3, line 65: If you do a `*object` when object is nil, go will panic. If there's an easy way to inject logs here or when the informers are created to have some information on what exactly is being gathered that could be good as well for troubleshooting purposes.

```suggestion
		if informer == nil {
```

retrieval/k8s/k8s_stats.go
```diff
@@ -1,82 +1,102 @@
...
+	daemonsetsInformer := factory.Apps().V1().DaemonSets().Informer()
+	deploymentsInformer := factory.Apps().V1().Deployments().Informer()
+	// Jobs
+	jobsInformer := factory.Batch().V1().Jobs().Informer()
+	// Cronjobs were introduced in k8s 1.21 so for older versions do not attempt to create an informer
+	var cronJobsInformer cache.SharedIndexInformer
+	if clusterVersion > 1.20 {
+		cronJobsInformer = factory.Batch().V1().CronJobs().Informer()
 	}
+	// closing this will kill all informers
```
jdhudson3, line 37: Ideally we should be closing this channel as well in order to allow the application/informer system and all related goroutines to cleanly shut down. If this could be either passed in from the main method or returned from here to a `defer close(stopCh)` that might be a bit safer.

retrieval/k8s/k8s_stats.go
```diff
@@ -1,82 +1,102 @@
...
+		}
+		_, err = datawriter.WriteString(string(data) + "\n")
+		if err != nil {
+			return errors.New("error: unable to write resource to file: " + resourceName)
+		}
 	}
 
-	return v1Sources, v1beta1Sources, v1AppSources
+	datawriter.Flush()
+	file.Close()
```
jdhudson3, line 99: Unhandled errors on both these lines

kubernetes/kubernetes.go
```diff
@@ -45,49 +46,52 @@ type ClusterVersion struct {
...
+	InClusterClient        raw.Client
+	msExportDirectory      *os.File
+	TLSClientConfig        rest.TLSClientConfig
+	Namespace              string
+	ScratchDir             string
+	NodeMetrics            EndpointMask
+	Informers              map[string]*cache.SharedIndexInformer
+	InformerResyncInterval int
+	ConcurrentPollers      int
+	ParseMetricData        bool
```
jdhudson3, line 88: Was the functionality for this flag re-implemented for informers?

testdata/e2e/e2e_helper.go
```diff
@@ -182,13 +215,98 @@ func shouldSkipFileCheck(fileType string, k8sMinorVersion int) bool {
...
+					return errors.New("cannot cast LabelSelectorMatchedResourceList")
+				}
+				typeStorage.Items = append(typeStorage.Items, *typedK8sObject)
+			case *LabelMapMatchedResource:
+				typeStorage, ok := storage.(*LabelMapMatchedResourceList)
+				if !ok {
+					return errors.New("cannot cast LabelMapMatchedResourceList")
+				}
+				typeStorage.Items = append(typeStorage.Items, *typedK8sObject)
+			}
```
jdhudson3, line 307: Default case with an error returned could be useful here.

retrieval/k8s/k8s_stats.go
daniel-spray: That is super interesting, because when I tried your suggestion, informer was a memory location. Then when I tested it with the dereference I got the nil value.

testdata/e2e/e2e_helper.go
daniel-spray: 👍 When I made changes in cake-services I forgot to update this. I also pulled the refFn call out of the loop here as well. TY

retrieval/k8s/k8s_stats.go
daniel-spray: I bumped the close to the CollectKubeMetrics function which should only ever complete in the case of the metrics-agent shutting down. let me know if this was incorrect implementation.

retrieval/k8s/k8s_stats.go
daniel-spray: Good catch

kubernetes/kubernetes.go
daniel-spray: no, going back and implementing these changes

daniel-spray: With most recent changes, yes this functionality got re-implemented :)

kubernetes/kubernetes_test.go
```diff
@@ -379,6 +410,77 @@ func TestCollectMetrics(t *testing.T) {
...
 		}
 	})
+	t.Run("Ensure collection occurs with parseMetrics is disabled"+
+		" ensure sensitive data is not stripped", func(t *testing.T) {
+
+		filepath.Walk(ka.msExportDirectory.Name(), func(path string, info os.FileInfo, err error) error {
+			if strings.Contains(info.Name(), "jsonl") {
+				if strings.Contains(info.Name(), "pods") {
+					// check if secrets were not stripped from pods if parseMetrics is false
+					in, _ := os.ReadFile(path)
```
jdhudson3, line 420: Not a huge deal since it's in a test file, but this could be bumped up to above the first if and the passed in error should be checked along with the return of os.ReadFile

kubernetes/kubernetes_test.go
```diff
@@ -379,6 +410,77 @@ func TestCollectMetrics(t *testing.T) {
 			}
 		}
 	})
+	t.Run("Ensure collection occurs with parseMetrics is disabled"+
+		" ensure sensitive data is not stripped", func(t *testing.T) {
+
+		filepath.Walk(ka.msExportDirectory.Name(), func(path string, info os.FileInfo, err error) error {
```
jdhudson3, line 416: Where is this sourcing data from? The output of a previous test?

kubernetes/kubernetes_test.go
```diff
@@ -519,5 +621,62 @@ func getMockInformers() map[string]*cache.SharedIndexInformer {
...
+	if err != nil {
+		t.Error(err)
+	}
+	deployments.Add(myDeployment)
+
 	return mockInformers
 }
+
+func startMockInformers(mockInformers map[string]*cache.SharedIndexInformer) {
+	informerStopCh := make(chan struct{})
```
jdhudson3, line 675: Would be nice to have a test that covers what happens if this channel is closed to ensure the informers wont hang the agent.

kubernetes/kubernetes_test.go
```diff
@@ -519,5 +621,62 @@ func getMockInformers() map[string]*cache.SharedIndexInformer {
...
+	if err != nil {
+		t.Error(err)
+	}
+	pods.Add(myPod)
+	// namespace also used in testing
+	namespaceData, _ := ioutil.ReadFile("../testdata/namespaces.jsonl")
+	var myNamespace *v1.Namespace
+	err = json.Unmarshal(namespaceData, &myNamespace)
+	if err != nil {
+		t.Error(err)
```
jdhudson3, line 659: Ideally should return an error in the cases and let the caller handle it.

kubernetes/nodecollection.go
```diff
@@ -136,7 +137,9 @@ func downloadNodeData(ctx context.Context, prefix string, config KubeAgentConfig
 			}
 			err := retrieveNodeData(nd, config, nodeSource, currentNode)
 			if err != nil {
+				m.Lock()
```
jdhudson3, line 140: Need a lock here too: https://github.com/cloudability/metrics-agent/pull/194/files#diff-7c1ca396f7b3ef69463ad472cf0120e4a854574ec4d837eea82655a8742a300dR127

Ideally would add tests for this as well. Can make a follow story for that if you want though.

retrieval/k8s/k8s_stats.go
```diff
@@ -1,82 +1,269 @@
...
 
-//GetK8sMetrics returns cloudabilty measurements retrieved from a given K8S Clientset
-func GetK8sMetrics(clusterHostURL string, clusterVersion float64, workDir *os.File, rawClient raw.Client) (err error) {
+func StartUpInformers(clientset kubernetes.Interface, clusterVersion float64,
+	resyncInterval int, stopCh chan struct{}) (map[string]*cache.SharedIndexInformer, error) {
+	factory := informers.NewSharedInformerFactory(clientset, time.Duration(resyncInterval)*time.Hour)
 
-	v1Sources, v1beta1Sources, v1AppSources := getk8sSourcePaths(clusterVersion)
+	// v1Sources
+	replicationControllerInformer := factory.Core().V1().ReplicationControllers().Informer()
```
jdhudson3, line 27: Minor: making `factory.Core().V1()` a variable would make this much more readable if that wont break anything

```suggestion
	replicationControllerInformer := factory.Core().V1().ReplicationControllers().Informer()
```

retrieval/k8s/k8s_stats.go
```diff
@@ -1,82 +1,269 @@
...
 
 	return err
 }
 
-func getk8sSourcePaths(clusterVersion float64) (v1Sources []string, v1beta1Sources []string, v1AppSources []string) {
-	commonSrcs := []string{
-		"replicasets",
-		"daemonsets",
-		"deployments",
+func getSanitizedK8sResource(k8Resource interface{}) interface{} {
```
jdhudson3, line 126: Does this function serve any purpose?

retrieval/k8s/k8s_stats.go
```diff
@@ -1,82 +1,269 @@
...
-		"pods",
-		"persistentvolumes",
-		"persistentvolumeclaims",
+	return to
+}
+
+func sanitizeMeta(objectMeta *metav1.ObjectMeta) {
+	objectMeta.ManagedFields = nil
+	if _, ok := objectMeta.Annotations[KubernetesLastAppliedConfig]; ok {
+		delete(objectMeta.Annotations, KubernetesLastAppliedConfig)
```
jdhudson3, line 229: Delete is a noop if the key doesn't exist, safe to call without the check.

kubernetes/kubernetes_test.go
daniel-spray: thanks, made the changes

kubernetes/kubernetes_test.go
daniel-spray: The source data comes from the mock informers that are created in the bottom of the test file.
https://github.com/cloudability/metrics-agent/blob/8916f8c744d2a3e96372aa7494e8ce2b8d483537/kubernetes/kubernetes_test.go#L574

daniel-spray: https://github.com/cloudability/metrics-agent/blob/3ab1bd801505e77b7fa2cc2f6c08b57afaa49d7a/testdata/pods.jsonl#L2-L1

kubernetes/kubernetes_test.go
daniel-spray: Oh so a test that checks to see if the agent crashes when the informers are closed?

kubernetes/kubernetes_test.go
daniel-spray: Sounds good, refactored

kubernetes/nodecollection.go
daniel-spray: Little confused, the comment says need a lock here too, and it is on the m.Lock() line for me.

retrieval/k8s/k8s_stats.go
daniel-spray: I can do that, but we would need 3 variables one for each resource "group/type" that we collect on:
- factory.Core().V1()
- factory.Batch().V1()
- factory.Apss().V1()
Examples:
	replicasetsInformer := factory.Apps().V1().ReplicaSets().Informer()
	jobsInformer := factory.Batch().V1().Jobs().Informer()
	nodesInformer := factory.Core().V1().Nodes().Informer()
If you still think it is worth it, I am fine adding it.

retrieval/k8s/k8s_stats.go
daniel-spray: nope, cleaned that up. Good call out

retrieval/k8s/k8s_stats.go
daniel-spray: I took this logic from your first implementation, does that mean I can also remove if's for sanitizePod's delete calls?

kubernetes/kubernetes_test.go
jdhudson3: In this case the scenario we want to avoid would be that we close the channel, but the informers goroutines don't stop.

kubernetes/nodecollection.go
jdhudson3: Weird, the link isn't working. On line 127 there needs to be a lock around this line:
```
			failedNodeList[currentNode.Name] = errors.New("provider ID for node does not exist. " +
"If this condition persists it will cause inconsistent cluster allocation")
```

retrieval/k8s/k8s_stats.go
jdhudson3: Yep. My bad for putting it there to begin with.

retrieval/k8s/k8s_stats.go
jdhudson3: Nah all good.

kubernetes/nodecollection.go
daniel-spray: thanks for clarifying, added the lock

retrieval/k8s/k8s_stats.go
daniel-spray: np just wanted to check. removed the if check

jdhudson3 (dismissed):

gizmo-rt:
@daniel-spray Could you help us merge the PR ?

PR #196, opened 2022-09-16
Enhance agent startup logging, more generous startup restrictions

jdhudson3:
#### What does this PR do?
Previously the agent would fail to start up in the case that the first "ready" node was unavailable or encountered an issue. 
This pr adds logging for all nodes on startup and changes the requirement to only fail if none are available.

Stick it in a cluster

Customer request to have more details information on reason for failure on startup as well as a more generous policy on per-node requirements.

Screen Shot 2022-09-16 at 10 34 26 AM

#### Developer Done List

- [ ] Tests Added/Updated
- [ ] Updated README.md

- [ ] Tests Pass Locally
- [ ] CI Build Passes

### DIFF
--- a/kubernetes/nodecollection.go
+++ b/kubernetes/nodecollection.go
@@ -29,7 +29,7 @@ func (err nodeError) Error() string {
 }
 
 const (
-	FatalNodeError = nodeError("unable to retrieve required set of node metrics via direct or proxy connection")
+	FatalNodeError = nodeError("unable to retrieve required metrics from any node via direct or proxy connection")
 )
 
 // NodeSource is an interface to get a list of Nodes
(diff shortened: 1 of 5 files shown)


kubernetes/nodecollection.go
```diff
@@ -348,34 +348,59 @@ func ensureNodeSource(ctx context.Context, config KubeAgentConfig) (KubeAgentCon
...
+				log.Warnf("Failed to connect to node [%s] directly with cause [%s]",
+					d.statsSummary(), err.Error())
+				failedDirect++
+			}
+			if success {
+				directlyConnected = true
+				directNodes++
+			}
+		}
+		if !directlyConnected {
```
bbylovas, line 376: Does the agent attempt to query through proxy if unable to connect directly? I think we should stick with whatever connection option was set in the config

kubernetes/nodecollection.go
```diff
@@ -348,34 +348,59 @@ func ensureNodeSource(ctx context.Context, config KubeAgentConfig) (KubeAgentCon
...
 		}
 	}
 
-	// test node connectivity via kube-proxy
-	p := setupProxyAPI(config.ClusterHostURL, firstNode.Name)
-	success, err := checkEndpointConnections(config, &config.HTTPClient, Proxy, p.statsSummary())
-	if err != nil {
-		return config, err
+	log.Infof("Of %d nodes, %d connected directly, %d connected via proxy, and %d could not be reached",
+		len(nodes), directNodes, proxyNodes, failedProxy)
```
bbylovas, line 391: I like logging the number of nodes that can't be reached. Don't think we need to include directNodes & proxyNodes since it should be an either/or type of situation

kubernetes/nodecollection.go
```diff
@@ -384,7 +409,7 @@ func checkEndpointConnections(config KubeAgentConfig, client *http.Client, metho
 	if err != nil {
 		return false, err
 	}
-	log.Infof("/stats/summary endpoint available via %s connection? %v", method, ns)
+	log.Infof("/stats/summary endpoint on node [%s] available via %s connection? %v", nodeStatSum, method, ns)
```
bbylovas, line 412: nodeStatSum is of the form "https://{{ ip }}:{{port}}/stats/summary" so we're repeating the endpoint name twice here

kubernetes/nodecollection.go
jdhudson3: The previous pattern was:
- Check if we are allowed to connect directly
- Attempt direct if so
- If not allowed to connect directly, or failed to connect directly, attempt proxy

This should follow the same pattern

kubernetes/nodecollection.go
jdhudson3: My thought was just to make it explicitly obvious how we were/were not able to connect, otherwise you have to look at the per-node logs above.

bbylovas (approved):
LGTM

apptiopzuo:
Have these changes tested? Have you tried with some nodes not reachable?

jdhudson3:
It's currently sitting in staging, we don't have any environments with invalid nodes.

kubernetes/nodecollection.go
```diff
@@ -415,17 +415,28 @@ func ensureNodeSource(ctx context.Context, config KubeAgentConfig) (KubeAgentCon
...
 	}
 
+	validateConfig(config, proxyNodes, directNodes)
 	return config, nil
 }
 
+func validateConfig(config KubeAgentConfig, proxyNodes, directNodes int32) {
+	if proxyNodes > 0 {
+		config.NodeMetrics.SetAvailability(NodeStatsSummaryEndpoint, Proxy, true)
+	} else if directNodes > 0 {
```
bbylovas, line 425: Should this be an `if` instead of an `else if`? Looks like SetAvailability can handle multiple connection methods

jdhudson3: It can take multiple but the downstream logic in the agent doesn't really support both being enabled. In `retrieveNodeData` if both were to be set, but a node only supported proxy the direct request would be attempted first, fail, and then return an error.

bbylovas (approved):
LGTM

PR #207, opened 2022-12-30
Health check on start up

daniel-spray:
#### What does this PR do?
Performs a health check by attempting to retrieve Presigned S3 URL on startup. This helps customers realize their agent is going to fail to upload metrics before waiting 10 minutes.

Also updates linter, fixes security vulnerability, and removes deprecated ioutil package.
Tested on test cluster
#### Developer Done List

- [ ] Tests Added/Updated
- [ ] Updated README.md

- [ ] Tests Pass Locally
- [ ] CI Build Passes


AnthonyApptio (dismissed):
LGTM

kubernetes/kubernetes.go
```diff
@@ -186,6 +187,41 @@ func CollectKubeMetrics(config KubeAgentConfig) {
...
+		ProxyAuth:     ka.OutboundProxyAuth,
+		ProxyInsecure: ka.OutboundProxyInsecure,
+	})
+	if err != nil {
+		log.Fatalf("error creating Cloudability Metric client: %v ", err)
+	}
+
+	metricSampleURL := apiEndpoint + "/metricsample"
+
+	file, err := os.Create("/tmp/temp.txt")
```
jdhudson3, line 206: https://pkg.go.dev/os#CreateTemp

Also is good for this.

Either way we should clean it up as a best practice.

```
defer os.Remove(file.Name())
```

kubernetes/heapster.go
```diff
@@ -88,6 +88,7 @@ func getHeapsterURL(ctx context.Context, clientset kubernetes.Interface, cluster
 	for _, pod := range pods.Items {
 		if strings.Contains(pod.Name, "heapster") {
 			URL.Host = clusterHostURL
+			//nolint staticcheck
```
jdhudson3, line 91: If this is complaining about it being a constant then we should probably just define one for it.

kubernetes/kubernetes.go
```diff
@@ -186,6 +187,41 @@ func CollectKubeMetrics(config KubeAgentConfig) {
...
+	if err != nil {
+		log.Warnf("Failed to create file temp.txt")
+	}
+
+	_, err = file.WriteString("Health Check")
+	if err != nil {
+		log.Warnf("Failed to write in file temp.txt")
+	}
+
+	_, _, err = cldyMetricClient.GetUploadURL(file, metricSampleURL, cldyVersion.VERSION, ka.clusterUID)
```
jdhudson3, line 216: Given that a customer could potentially have the upload domain blocked it could be kind of a convenient check to have it upload the tmp file as well assuming that it overwrites seamlessly.

I don't know if we've seen that specific issue occur previously, but I know we have seen general connectivity on the actual upload process several times.

test/random.go
```diff
@@ -42,7 +42,7 @@ func SecureRandomAlphaString(length int) string {
 // SecureRandomBytes returns the requested number of bytes using crypto/rand
 func SecureRandomBytes(length int) []byte {
 	var randomBytes = make([]byte, length)
-	_, err := rand.Read(randomBytes)
+	_, err := r.Read(randomBytes)
```
jdhudson3, line 45: crypto.rand should likely be kept. math.rand is less random

kubernetes/kubernetes.go
```diff
@@ -186,6 +187,41 @@ func CollectKubeMetrics(config KubeAgentConfig) {
...
+	})
+	if err != nil {
+		log.Fatalf("error creating Cloudability Metric client: %v ", err)
+	}
+
+	metricSampleURL := apiEndpoint + "/metricsample"
+
+	file, err := os.Create("/tmp/temp.txt")
+	if err != nil {
+		log.Warnf("Failed to create file temp.txt")
```
jdhudson3, line 208: If you don't return from here, line 211 will panic. Each of these return should probably trigger a failure to start the agent as they're basic file io steps.

jdhudson3: Returning an error from here and logging it from the caller might be wise as well, would be easier to force an agent restart that way as well.

kubernetes/kubernetes.go
daniel-spray: Good call, added defer line

kubernetes/heapster.go
daniel-spray: This complains about the pod.selfLink saying it is deprecated. I just no linted because this is heapster code that we no longer use.
The linter:
kubernetes/heapster.go:91:15: SA1019: pod.SelfLink is deprecated: selfLink is a legacy read-only field that is no longer populated by the system. +optional (staticcheck)

kubernetes/kubernetes.go
daniel-spray: good catch, making changes, thanks!

test/random.go
jdhudson3: Weird, docs say the opposite: 

```This package's outputs might be easily predictable regardless of how it's seeded. For random numbers suitable for security-sensitive work, see the crypto/rand package. 
```
seems like this is only for tests though, so either is prob fine.

jdhudson3: Ah, yeah the linter is saying that math.rand is weaker. It just will only complain if both are imported.

kubernetes/kubernetes.go
daniel-spray: Yeah we have seen the issue where a customer can get the url but not upload however it is very rare. We also discussed this and don't want to send unnecessary file uploads to our s3. Could be something we revisit depending on the number of issues where uploading sees error when s3 url is fine.

PR #211, opened 2023-03-02
Better Upload Logs

daniel-spray:
#### What does this PR do?
Improves logging for troubleshooting upload errors for the following scenarios.

When in INFO mode:
- When successful, for both the presigned url request AND the S3 Put request. Print the aws request-id along with accurate messaging. 
- When failing, for either presigned url request OR the S3 Put request. Print the aws request-id along with accurate failure messaging/error code.

When in DEBUG mode:
- When successful, follow same pattern as INFO success.
- When failing, for either presigned url request OR the S3 Put request. Add an extra log that will dump the entire HTTP request which includes aws request-id.

Deployed on a test cluster and tested various scenarios
Nope
#### Developer Done List

- [x] Tests Added/Updated
- [x] Updated README.md

- [ ] Tests Pass Locally
- [ ] CI Build Passes

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -136,7 +136,7 @@ func NewHTTPMetricClient(cfg Configuration) (MetricClient, error) {
 type MetricClient interface {
 	SendMeasurement(measurements []measurement.Measurement) error
 	SendMetricSample(*os.File, string, string) error
-	GetUploadURL(*os.File, string, string, string) (string, string, error)
+	GetUploadURL(*os.File, string, string, string, int) (string, string, error)
 }
 
 type httpMetricClient struct {
@@ -214,11 +214,6 @@ func (c httpMetricClient) SendMetricSample(metricSampleFile *os.File, agentVersi
 
 	defer util.SafeClose(resp.Body.Close, &rerr)
 
-	responseDump, err := httputil.DumpResponse(resp, true)
-	if err != nil {
-		log.Errorln(err)
-	}
-
 	if resp.StatusCode != http.StatusOK {
 		return fmt.Errorf("Request received %v response", resp.StatusCode)
 	}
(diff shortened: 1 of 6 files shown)


client/client.go
```diff
@@ -253,30 +245,36 @@ func (c httpMetricClient) retryWithBackoff(
...
 		if err != nil && strings.Contains(err.Error(), "Client.Timeout exceeded") {
 			time.Sleep(getSleepDuration(i))
+			log.Infof("Put S3 Retry %d: Failed to put data to S3, X-Amzn-Requestid: %s", i, awsRequestID)
+			log.Debugln(string(responseDump))
 			continue
 		}
 
 		if resp == nil {
+			log.Infof("Put S3 Retry %d: Failed to put data to S3, X-Amzn-Requestid: %s", i, awsRequestID)
+			log.Debugln(string(responseDump))
```
jdhudson3, line 277: Will these debug statements always be emitted, or is the default log level info?

daniel-spray: The default log level is INFO. 
These messages only display if they set the CLOUDABILITY_LOG_LEVEL env variable to "DEBUG"

jdhudson3: Given that these logs are only emitted in error cases I think we might as well just always log them.

Otherwise during the support process when we would need them someone will have to go configure the agent and wait for the error to reoccur.

daniel-spray: Updated so we print the Status Message (ex: 403 Forbidden) in INFO mode. 
Leaving the whole log dump to DEBUG to prevent sensitive information being dumped in logs in INFO mode

client/client.go
```diff
@@ -253,30 +245,39 @@ func (c httpMetricClient) retryWithBackoff(
...
-		if c.verbose {
-			reponseDump, err := httputil.DumpResponse(resp, true)
-			if err != nil {
-				log.Errorln(err)
+		if resp != nil {
+			awsRequestID = resp.Header.Get("X-Amz-Request-Id")
+			statusMessage = resp.Status
+			responseDump, dumpErr = httputil.DumpResponse(resp, true)
+			if dumpErr != nil {
+				log.Errorln(dumpErr)
```
jdhudson3, line 264: I'm not sold on keeping the `continue` in here, if the request succeeds then I don't think we care if the response failed

client/client.go
jdhudson3: The debug in this block will never contain anything since it's based off the data in resp.

client/client.go
```diff
@@ -253,30 +245,39 @@ func (c httpMetricClient) retryWithBackoff(
...
 				continue
 			}
-			log.Infoln(string(reponseDump))
 		}
 
 		if err != nil && strings.Contains(err.Error(), "Client.Timeout exceeded") {
 			time.Sleep(getSleepDuration(i))
+			log.Infof("Put S3 Retry %d: Failed to put data to S3, Status: %s X-Amzn-Requestid: %s", i,
```
jdhudson3, line 271:
```suggestion
			log.Infof("Put S3 Retry %d: Failed to put data to S3 due to request timeout, Status: %s X-Amzn-Requestid: %s", i,
```

jdhudson3: Seems like all of these should be at least warn level

daniel-spray: 👍

client/client.go
daniel-spray: I also think the infof the statusMessage and the awsRequestID will also be nil here too. I can remove them, but in this case I am not sure what the error message should be? 
"Put S3 Retry %d: Failed to put data to S3. Nil Response". 
That log line does not give us anything super helpful...

PR #233, opened 2023-09-26
Add CLOUDABILITY_UPLOAD_REGION Env Variable

daniel-spray:
#### What does this PR do?
Adds a new environment variable CLOUDABILITY_UPLOAD_REGION
Supported value:
- us-west-2: Uploads to Apptio's US S3
For context: There was two global variables for for the metrics collection endpoint in different packages. I removed one from a package and imported it to the other so we don't need to store duplicate global variables which could cause confusion.
Ran tests and deployed onto test cluster.
Nope
#### Developer Done List

- [x] Tests Added/Updated
- [x] Updated README.md

- [ ] Tests Pass Locally
- [ ] CI Build Passes

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -407,8 +407,8 @@ func GetB64MD5Hash(name string) (b64Hash string, rerr error) {
 	return base64.StdEncoding.EncodeToString(h.Sum(nil)), err
 }
 
-// GetUploadURLRegion returns the correct base url depending on the env variable CLOUDABILITY_UPLOAD_REGION. If value is not
-// supported, default to us-west-2 (original) URL
+// GetUploadURLRegion returns the correct base url depending on the env variable CLOUDABILITY_UPLOAD_REGION.
+// If value is not supported, default to us-west-2 (original) URL
 func GetUploadURLRegion(region string) string {
 	switch region {
 	case "eu-central-1":
--- a/client/client.go
+++ b/client/client.go
@@ -78,7 +78,7 @@ func NewHTTPMetricClient(cfg Configuration) (MetricClient, error) {
 		if cfg.Verbose {
 			log.Infof("Using default baseURL of %v", DefaultBaseURL)
 		}
-		cfg.BaseURL = GetUploadURLRegion(cfg.Region)
+		cfg.BaseURL = GetUploadURLByRegion(cfg.Region)
 	}
 	if cfg.MaxRetries <= 0 {
 		if cfg.Verbose {
@@ -407,9 +407,9 @@ func GetB64MD5Hash(name string) (b64Hash string, rerr error) {
 	return base64.StdEncoding.EncodeToString(h.Sum(nil)), err
 }
 
-// GetUploadURLRegion returns the correct base url depending on the env variable CLOUDABILITY_UPLOAD_REGION.
+// GetUploadURLByRegion returns the correct base url depending on the env variable CLOUDABILITY_UPLOAD_REGION.
 // If value is not supported, default to us-west-2 (original) URL
-func GetUploadURLRegion(region string) string {
+func GetUploadURLByRegion(region string) string {
 	switch region {
 	case "eu-central-1":
 		return EUBaseURL
--- a/client/client.go
+++ b/client/client.go
@@ -30,7 +30,7 @@ import (
 
 //nolint gosec
 
-const DefaultBaseURL = "https://metrics-collector.cloudability.com"
+const DefaultBaseURL string = "https://metrics-collector.cloudability.com"
 const EUBaseURL string = "https://metrics-collector-eu.cloudability.com"
 const defaultTimeout = 1 * time.Minute
 const defaultMaxRetries = 5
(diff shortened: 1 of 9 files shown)


client/client.go
```diff
@@ -30,7 +30,8 @@ import (
 
 //nolint gosec
 
-const defaultBaseURL = "https://metrics-collector.cloudability.com"
+const DefaultBaseURL = "https://metrics-collector.cloudability.com"
+const EUBaseURL string = "https://metrics-collector-eu.cloudability.com"
 const defaultTimeout = 1 * time.Minute
 const defaultMaxRetries = 5
```
jyin-apptio, line 36: Consistency? no `string` and first character lowercase?

client/client.go
```diff
@@ -452,3 +454,21 @@ func GetB64MD5Hash(name string) (b64Hash string, rerr error) {
...
+		return EUBaseURL
+	case "eu-west-1":
+		return EUBaseURL
+	case "us-west-2":
+		return DefaultBaseURL
+	case "us-east-1":
+		return DefaultBaseURL
+	default:
+		log.Warnf("Region %s is not supported. Defaulting to us-west-2 region.", region)
+		return DefaultBaseURL
```
jyin-apptio, line 472: Why do we need 4 regions if we only have 2 URL?

kubernetes/kubernetes.go
```diff
@@ -219,12 +218,13 @@ func performConnectionChecks(ka *KubeAgentConfig) error {
...
 		ProxyInsecure: ka.OutboundProxyInsecure,
 		Timeout:       time.Duration(ka.HTTPSTimeout) * time.Second,
+		Region:        ka.UploadRegion,
 	})
 	if err != nil {
 		return errors.New("error creating Cloudability Metric client in connectivity test")
 	}
 
-	metricSampleURL := apiEndpoint + "/metricsample"
+	metricSampleURL := client.GetUploadURL(ka.UploadRegion) + "/metricsample"
```
jyin-apptio, line 227: Nitpick, could we move `metricsample` within the GetUploadURL method...

client/client.go
daniel-spray: I did not mean to drop the string type declaration, I will add it back.

The reason to capitalize d -> D is we now use the variable outside of the client package so it needs to be a public var.

kubernetes/kubernetes.go
daniel-spray: The upload URL is logged without this /metricsample in an error log. Not sure if we care that this would change but I figured just leave as is to reduce change radius

jyin-apptio: Is there any concern to log out the `metricsample` as well.
The idea for this comment is to have unified func to generate the URL to minimize the risk that we test URL_A here and actual getURL is using URL_B because it is generated in different way

daniel-spray: No concern with logging /metricsample. My concern is with extensiblity. It looks like this url was originally committed to support several endpoiints but ended up only needing /metricsample. For example there is a SendMeasurement function that hits /metrics https://github.com/cloudability/metrics-agent/blob/2c8c5421b0ceae0e556398d0560af641448d8ff0/client/client.go#L157

It seems unused expect by test files and is from the original commit. I can add the /metricsample to the base url since I doubt we have any need for other endpoints in the MA

daniel-spray: I feel changing this base url is not as simple as a 1 line change as some testing structure is based on it. So I don't really see the value. If it is a blocker to merge then I will rewrite the test files to support the base url having `/metricsample` if needed.

daniel-spray: I do get your point though that this is still two separate url generations. Would a method like getMetricsSampleURL() that returns a string with the entire url be better in your opinion. So that we can do something like
`metricSampleURL := getMetricsSampleURL(client.GetUploadURL(ka.UploadRegion))` in both places so we know generation is the same as we use the same method.

jyin-apptio: For test, could we just update baseURL in the code to match?
https://github.com/cloudability/metrics-agent/blob/78faf44041d249581fec5d4361b8d7dd419a4f6c/client/client_test.go#L111
And current test seems doesn't cover the region specific logic and we only care about the URI suffix is `/metrcisample` , which seems a gap. 

I would in favor of making our code cleaner rather than worry about the extensibility to a non-valid use case (SendMeasurement).

client/client.go
```diff
@@ -452,3 +406,17 @@ func GetB64MD5Hash(name string) (b64Hash string, rerr error) {
 
 	return base64.StdEncoding.EncodeToString(h.Sum(nil)), err
 }
+
+// GetUploadURLRegion returns the correct base url depending on the env variable CLOUDABILITY_UPLOAD_REGION.
+// If value is not supported, default to us-west-2 (original) URL
+func GetUploadURLRegion(region string) string {
```
jyin-apptio, line 412: Nit, s/GetUploadURLRegion/GetUploadURL. The previous name sounds like return region value

jyin-apptio (approved):
LGTM

PR #275, opened 2024-10-14
azure blob storage upload support

AnthonyApptio:
#### What does this PR do?
Adds support for uploading to a custom Azure blob storage

Created an image based on this build and applied the image to a metrics-agent yaml. Deployed that metrics-agent to a test cluster.
Set all custom azure env vars with the correct values.
Confirm uploads to the Azure blob container name provided.

#### Developer Done List

- [ ] Tests Added/Updated
- [ ] Updated README.md

- [ ] Tests Pass Locally
- [ ] CI Build Passes


README.md
```diff
@@ -90,6 +95,11 @@ Flags:
       --upload_region                            The region the metrics-agent will upload data to. (default `us-west-2`)
       --custom_s3_bucket string                  A custom S3 bucket the metrics-agent will upload data to. - Optional
       --custom_s3_region                         The AWS region that the custom s3 bucket is created.
+      --custom_azure_blob_container_name string  A custom Azure blob container name the metrics-agent will upload data to. - Optional
+      --custom_azure_blob_url                    A custom Azure blob url metrics-agent uses to create the Azure client.
+      --custom_azure_tenant_id                   A custom Azure Service principal tenant id the metrics-agent uses to create the Azure client.
```
daniel-spray, line 100: this documentation I think is a bit messy already but shouldn't all these --custom_azure flags have `- Optional`

kubernetes/kubernetes.go
```diff
@@ -492,13 +526,61 @@ func (ka KubeAgentConfig) sendMetricsToCustomS3(metricSample *os.File) {
...
 // /production/data/metrics-agent/<YYYY>/<MM>/<DD>/<CLUSTER_UID>/<CLUSTER_UID>-<YYYYMMDD>-<HH>-<MM>.tgz
-func generateSampleKey(clusterUID string) string {
+// example of azure location format
+// production/data/metrics-agent/<YYYY>/<MM>/<DD>/<CLUSTER_UID>/<CLUSTER_UID>-<YYYYMMDD>-<HH>-<MM>.tgz
+func generateSampleKey(clusterUID, provider string) string {
 	currentTime := time.Now()
 	year, month, day := currentTime.Date()
 	hour := currentTime.Hour()
 	minute := currentTime.Minute()
+	if provider == "aws" {
```
daniel-spray, line 576: Seems like the sampleKey just changes by 1 leading `/`. 
Might be a bit cleaner to just remove the leading `/` if provider is `azure`.
So this function does not need to duplicate: 
`
return fmt.Sprintf("production/data/metrics-agent/%d/%02d/%02d/%s/%s-%s-%02d-%02d.tgz", year,
			int(month), day, clusterUID, clusterUID, currentTime.Format("20060102"), hour, minute)
`
3 times.

kubernetes/kubernetes.go
```diff
@@ -492,13 +526,61 @@ func (ka KubeAgentConfig) sendMetricsToCustomS3(metricSample *os.File) {
 	}
 }
 
-// generateSampleKey creates a key (location) for s3 to upload the sample to. Example of s3 location format
+func (ka KubeAgentConfig) sendMetricsToCustomBlob(metricSample *os.File) {
+	os.Setenv("AZURE_TENANT_ID", ka.CustomAzureTenantID)
```
daniel-spray, line 530: is setting the environment variables the best practice here?
Can they not just be injected into the cred/client?

kubernetes/kubernetes.go
```diff
@@ -136,6 +143,7 @@ func CollectKubeMetrics(config KubeAgentConfig) {
 	kubeAgent := newKubeAgent(ctx, config)
 
 	customS3Mode := isCustomS3UploadEnvsSet(&kubeAgent)
+	customAzureMode := isCustomAzureUploadEnvsSet(&kubeAgent)
```
daniel-spray, line 146: Now that we have two modes, I wonder if it makes sense to make an `agentUploadMode` variable instead of managing separate vars and passing double parameters. Especially if other locations are needed to be supported in the future GKE, OKE...
Basically could have a map of `uploadModes` that could determine future actions

README.md
AnthonyApptio: i'll fix them all while here

kubernetes/kubernetes.go
AnthonyApptio: couldn't find a good example not setting the envs will double check w/ John tomorrow about this.

go.mod
```diff
@@ -33,18 +38,20 @@ require (
...
 	github.com/google/safetext v0.0.0-20220905092116-b49f7bc46da2 // indirect
-	github.com/google/uuid v1.3.0 // indirect
+	github.com/google/uuid v1.6.0 // indirect
 	github.com/hashicorp/hcl v1.0.0 // indirect
 	github.com/imdario/mergo v0.3.13 // indirect
 	github.com/inconshreveable/mousetrap v1.0.1 // indirect
 	github.com/jmespath/go-jmespath v0.4.0 // indirect
 	github.com/josharian/intern v1.0.0 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
+	github.com/kylelemons/godebug v1.1.0 // indirect
```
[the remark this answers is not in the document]
AnthonyApptio: this is part of the azure stuff being brought in

kubernetes/kubernetes.go
AnthonyApptio: Updated to now use: https://github.com/Azure/azure-sdk-for-go/blob/sdk/azidentity/v1.8.0/sdk/azidentity/client_secret_credential.go#L47 no longer setting the envs that way.

kubernetes/kubernetes.go
```diff
@@ -234,7 +256,32 @@ func isCustomS3UploadEnvsSet(ka *KubeAgentConfig) bool {
...
 		"Will upload collected metrics to %s in the aws region %s", ka.CustomS3UploadBucket, ka.CustomS3Region)
-	return true
+	return true, "aws"
+}
+
+// isCustomAzureUploadEnvsSet checks to see if the agent has custom Azure credentials needed for uploads
+// if these vars are found, set the env vars needed for the azure client. If not set, default upload to Apptio S3
+func isCustomAzureUploadEnvsSet(ka *KubeAgentConfig) (bool, string) {
+	if ka.CustomAzureUploadBlobContainerName == "" && ka.CustomAzureBlobURL == "" && ka.CustomAzureClientID == "" &&
+		ka.CustomAzureClientSecret == "" && ka.CustomAzureTenantID == "" {
```
jdhudson3, line 266: Could break out the giant equals blob and set it to a bool rather than duplicating it.

kubernetes/kubernetes.go
```diff
@@ -234,7 +256,32 @@ func isCustomS3UploadEnvsSet(ka *KubeAgentConfig) bool {
...
+		if ka.APIKey == "" {
+			log.Fatalf("Invalid agent configuration. CLOUDABILITY_API_KEY is required " +
+				"when not using CLOUDABILITY_CUSTOM_AZURE_BLOB env vars.")
+		}
+		return false, ""
+	}
+	if ka.CustomAzureUploadBlobContainerName == "" || ka.CustomAzureBlobURL == "" || ka.CustomAzureClientID == "" ||
+		ka.CustomAzureClientSecret == "" || ka.CustomAzureTenantID == "" {
+		log.Fatalf("Invalid agent configuration. Detected only one of the six required environment variables "+
```
jdhudson3, line 275: Six required vars or five?

kubernetes/kubernetes.go
```diff
@@ -234,7 +256,32 @@ func isCustomS3UploadEnvsSet(ka *KubeAgentConfig) bool {
...
+			"to run in custom Azure upload mode. CLOUDABILITY_CUSTOM_AZURE_BLOB_CONTAINER_NAME is set to %s. "+
+			"CLOUDABILITY_CUSTOM_AZURE_BLOB_URL is set to %s CLOUDABILITY_CUSTOM_AZURE_TENANT_ID is set to %s. "+
+			"CLOUDABILITY_CUSTOM_AZURE_CLIENT_ID is set to %s. CLOUDABILITY_CUSTOM_AZURE_CLIENT_SECRET is set to %s.",
+			ka.CustomAzureUploadBlobContainerName, ka.CustomAzureBlobURL, ka.CustomAzureClientID, ka.CustomAzureTenantID,
+			ka.CustomAzureClientSecret)
```
jdhudson3, line 280: Don't log out secrets or other sensitive data. Can check that they != "" and log that they are set if you want context for them.

kubernetes/kubernetes.go
```diff
@@ -443,11 +490,17 @@ func (ka KubeAgentConfig) sendMetrics(metricSample *os.File) {
 	}
 }
 
-func (ka KubeAgentConfig) sendMetricsBasedOnUploadMode(customS3Mode bool, metricSample *os.File) {
-	if customS3Mode {
+// sendMetricsBasedOnUploadMode checks the agentProvider for a custom upload config. If none are found it will use the
+// default upload mode. This will allow for more custom provider upload options in the future.
+func (ka KubeAgentConfig) sendMetricsBasedOnUploadMode(agentProvider string, metricSample *os.File) {
+	switch agentProvider {
+	case "s3":
```
jdhudson3, line 497: Looks like this will break custom s3 uploads. Seems like `agentProvider` is `aws` now currently, should swap for a const or enum on upload types.

kubernetes/kubernetes.go
```diff
@@ -217,15 +226,28 @@ func CollectKubeMetrics(config KubeAgentConfig) {
...
+		return isCustomUpload, agentMode
+	}
+	isCustomUpload, agentMode = isCustomAzureUploadEnvsSet(ka)
+	return isCustomUpload, agentMode
+}
+
 // isCustomS3UploadEnvsSet checks to see if the agent has a custom S3 location and S3 region to upload to
 // if both these variables are not set, default upload to Apptio S3
-func isCustomS3UploadEnvsSet(ka *KubeAgentConfig) bool {
+func isCustomS3UploadEnvsSet(ka *KubeAgentConfig) (bool, string) {
```
jdhudson3, line 244: String return here is unnecessary imo, could consolidate this into `isCustomUploadEnvsSet`. Alternatively an enum or three consts could satisfy the requirements of both vars.

isCustomS3Upload := isCustomS3UploadEnvsSet(ka)
	if isCustomS3Upload {
		return true, aws
	}
	isCustomAzureUpload := isCustomAzureUploadEnvsSet(ka)
	if isCustomAzureUpload {
		return true, azure
	}
	return false,""

kubernetes/kubernetes.go
```diff
@@ -492,13 +545,55 @@ func (ka KubeAgentConfig) sendMetricsToCustomS3(metricSample *os.File) {
...
 // /production/data/metrics-agent/<YYYY>/<MM>/<DD>/<CLUSTER_UID>/<CLUSTER_UID>-<YYYYMMDD>-<HH>-<MM>.tgz
-func generateSampleKey(clusterUID string) string {
+// example of azure location format
+// production/data/metrics-agent/<YYYY>/<MM>/<DD>/<CLUSTER_UID>/<CLUSTER_UID>-<YYYYMMDD>-<HH>-<MM>.tgz
+func generateSampleKey(clusterUID, provider string) string {
 	currentTime := time.Now()
 	year, month, day := currentTime.Date()
 	hour := currentTime.Hour()
 	minute := currentTime.Minute()
+	if provider == "azure" {
```
jdhudson3, line 593: If the only difference is the slash it'd be cleaner to just prepend that in the s3 case rather than duplicate the format. Either via a direct string concat or a `pathPrefix` var that is `""` for azure

kubernetes/kubernetes.go
```diff
@@ -833,6 +928,11 @@ func createAgentStatusMetric(workDir *os.File, config KubeAgentConfig, sampleSta
 	m.Values["upload_region"] = config.UploadRegion
 	m.Values["custom_s3_bucket"] = config.CustomS3UploadBucket
 	m.Values["custom_s3_region"] = config.CustomS3Region
+	m.Values["custom_azure_blob_container_name"] = config.CustomAzureUploadBlobContainerName
+	m.Values["custom_azure_blob_url"] = config.CustomAzureBlobURL
+	m.Values["custom_azure_tenant_id"] = config.CustomAzureTenantID
+	m.Values["custom_azure_client_id"] = config.CustomAzureClientID
+	m.Values["custom_azure_client_secret"] = config.CustomAzureClientSecret
```
jdhudson3, line 935:
```suggestion
```

We don't ever need this information, should be private to the client.

cmd/kubernetesCmd.go
```diff
@@ -150,6 +150,37 @@ func init() {
 		"",
 		"The AWS region that the custom s3 bucket is in",
 	)
+	kubernetesCmd.PersistentFlags().StringVar(
+		&config.CustomAzureUploadBlobContainerName,
+		"custom_azure_blob_container_name",
+		"",
+		"The Azure blob container name the metrics-agent will upload data to. Default is an empty string "+
+			"which will not upload to custom azure location",
```
jdhudson3, line 158:
```suggestion
		"The Azure blob container name the metrics-agent will upload data to.",
```

Providing information about a default that doesn't work is confusing imo

kubernetes/kubernetes.go
```diff
@@ -492,13 +545,55 @@ func (ka KubeAgentConfig) sendMetricsToCustomS3(metricSample *os.File) {
...
 
-// generateSampleKey creates a key (location) for s3 to upload the sample to. Example of s3 location format
+func (ka KubeAgentConfig) sendMetricsToCustomBlob(metricSample *os.File) {
+	cred, err := azidentity.NewClientSecretCredential(ka.CustomAzureTenantID, ka.CustomAzureClientID,
+		ka.CustomAzureClientSecret, nil)
+	if err != nil {
+		log.Fatalf("Could not establish Azure credentials, "+
+			"ensure Azure environment variables are set correctly: %s", err)
+	}
+	azureClient, err := azblob.NewClient(ka.CustomAzureBlobURL, cred, nil)
```
jdhudson3, line 555: Worth adding settings for a retry here imo. We do 3 for S3

kubernetes/kubernetes.go
```diff
@@ -492,13 +545,55 @@ func (ka KubeAgentConfig) sendMetricsToCustomS3(metricSample *os.File) {
...
+	file, err := os.Open(metricSample.Name())
+	if err != nil {
+		log.Fatalf("Unable to open metric sample file %v", err)
+	}
+	defer file.Close()
+	key := generateSampleKey(ka.clusterUID, azure)
+	_, err = client.UploadFile(context.Background(), ka.CustomAzureUploadBlobContainerName, key, file, nil)
+	if err != nil {
+		log.Fatalf("Failed to put Object to custom Azure blob with error: %s", err)
+	} else {
```
jdhudson3, line 573: Redundant else condition

kubernetes/kubernetes.go
```diff
@@ -492,13 +545,55 @@ func (ka KubeAgentConfig) sendMetricsToCustomS3(metricSample *os.File) {
...
+		sn := strings.Split(metricSample.Name(), "/")
```
jdhudson3, line 574: Why bother trimming off the suffix? Seems more reasonable to keep the full path since that's what would be in the blob storage

kubernetes/kubernetes.go
```diff
@@ -492,13 +545,55 @@ func (ka KubeAgentConfig) sendMetricsToCustomS3(metricSample *os.File) {
...
+}
+
+func (ka KubeAgentConfig) uploadBlob(client *azblob.Client, metricSample *os.File) {
+	file, err := os.Open(metricSample.Name())
+	if err != nil {
+		log.Fatalf("Unable to open metric sample file %v", err)
+	}
+	defer file.Close()
+	key := generateSampleKey(ka.clusterUID, azure)
+	_, err = client.UploadFile(context.Background(), ka.CustomAzureUploadBlobContainerName, key, file, nil)
```
jdhudson3, line 570: Best practice would be to pass + use the parent context here since we have one defined already.

kubernetes/kubernetes.go
```diff
@@ -492,13 +545,55 @@ func (ka KubeAgentConfig) sendMetricsToCustomS3(metricSample *os.File) {
...
+		log.Warnf("Warning: Unable to cleanup after metric sample upload: %v", err)
+	}
+}
+
+// generateSampleKey creates a key (location) for s3 or azure to upload the sample to. Example of s3 location format
 // /production/data/metrics-agent/<YYYY>/<MM>/<DD>/<CLUSTER_UID>/<CLUSTER_UID>-<YYYYMMDD>-<HH>-<MM>.tgz
-func generateSampleKey(clusterUID string) string {
+// example of azure location format
+// production/data/metrics-agent/<YYYY>/<MM>/<DD>/<CLUSTER_UID>/<CLUSTER_UID>-<YYYYMMDD>-<HH>-<MM>.tgz
+func generateSampleKey(clusterUID, provider string) string {
```
jdhudson3, line 588: Existing code, but 
`currentTime := time.Now()` should likely be `currentTime := time.Now().UTC()`

kubernetes/kubernetes.go
```diff
@@ -492,13 +545,55 @@ func (ka KubeAgentConfig) sendMetricsToCustomS3(metricSample *os.File) {
...
+	}
+	ka.uploadBlob(azureClient, metricSample)
+}
+
+func (ka KubeAgentConfig) uploadBlob(client *azblob.Client, metricSample *os.File) {
+	file, err := os.Open(metricSample.Name())
+	if err != nil {
+		log.Fatalf("Unable to open metric sample file %v", err)
+	}
+	defer file.Close()
```
jdhudson3, line 568: Defer close but inline removal seems like it could cause a problem? Should probably defer them both in the order that the file is closed then removed.

kubernetes/kubernetes.go
AnthonyApptio: can't break them out as the first check is and operators and the second check is or operators

kubernetes/kubernetes.go
AnthonyApptio: This is to match the same formatting as the default mode: https://github.com/cloudability/metrics-agent/blob/220145f405967b0641d31553625152538d7c3600/kubernetes/kubernetes.go#L617

kubernetes/kubernetes.go
AnthonyApptio: My only argument of not changing this is we do appear to have 2 customers using the custom S3 upload feature. If we change to UTC that could effect their upload times. If the customer is using the timestamp for anything i don't think we can just change it. Thoughts?

jdhudson3: We can verify that this will not impact any customers using it

kubernetes/kubernetes.go
```diff
@@ -229,25 +233,26 @@ func CollectKubeMetrics(config KubeAgentConfig) {
...
+	return false, ""
 }
 
 // isCustomS3UploadEnvsSet checks to see if the agent has a custom S3 location and S3 region to upload to
 // if both these variables are not set, default upload to Apptio S3
-func isCustomS3UploadEnvsSet(ka *KubeAgentConfig) (bool, string) {
+func isCustomS3UploadEnvsSet(ka *KubeAgentConfig) bool {
 	if ka.CustomS3Region == "" && ka.CustomS3UploadBucket == "" {
 		if ka.APIKey == "" {
 			log.Fatalf("Invalid agent configuration. CLOUDABILITY_API_KEY is required " +
```
jdhudson3, line 252: Have you tested the agent in the azure configuration? Seems like this would just crash.

I don't think a bunch of log.Fatal's is a great handling method for this when multiple configurations need to be interpreted.

kubernetes/kubernetes.go
```diff
@@ -545,57 +556,67 @@ func (ka KubeAgentConfig) sendMetricsToCustomS3(metricSample *os.File) {
...
-	err = os.Remove(metricSample.Name())
-	if err != nil {
-		log.Warnf("Warning: Unable to cleanup after metric sample upload: %v", err)
 	}
+
+	sn := strings.Split(metricSample.Name(), "/")
+	log.Infof("Exported metric sample %s to custom Azure blob: %s",
+		strings.TrimSuffix(sn[len(sn)-1], ".tgz"), ka.CustomAzureUploadBlobContainerName)
+
+	defer func(name string) {
```
jdhudson3, line 597: Defers are operated on as a LIFO stack. In this case it will still remove before closing.

kubernetes/kubernetes.go
AnthonyApptio: Yes tested the agent with the azure configs. When the aws config isn't set both ka.CustomS3Region and ka.CustomS3UploadBucket are set to "" the API key must be set so it will return false. Then it will check the azure config.

The log.Fatal's seem to be necessary to break the process when a custom upload config is partially set but not all values are there so it will not revert to the default upload. Also the log will show the customer which env config is not set correctly.

kubernetes/kubernetes.go
AnthonyApptio: verified adding this UTC will not be a breaking change for current customers

kubernetes/kubernetes.go
AnthonyApptio: Added a defer function and both close and remove are now handled LIFO within that function.

util/util.go
```diff
@@ -288,6 +288,23 @@ func SafeClose(closer func() error, err *error) {
 	}
 }
 
+// SafeCloseDeferLogger takes a file and runs the deferred closing of both. Used for custom agent uploads
+func SafeCloseDeferLogger(file *os.File) {
+	defer func(name string) {
```
jdhudson3, line 293: Deferring things within a method doesn't do much for you.

In this case consider having the two operations run normally (close -> defer), and defer this function from the caller.

jdhudson3 (dismissed):

charts/metrics-agent/Chart.yaml
```diff
@@ -18,4 +18,4 @@ version: 2.11.36
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
-appVersion: 2.11.36
\ No newline at end of file
+appVersion: 2.11.36
```
daniel-spray, line 21: this release needs to be 2.11.37 as .36 was already released

kubernetes/kubernetes.go
```diff
@@ -505,8 +496,15 @@ func (ka KubeAgentConfig) sendMetricsBasedOnUploadMode(ctx context.Context, agen
 		log.Infof("Uploading Metrics to Custom S3 Bucket %s", ka.CustomS3UploadBucket)
 		go ka.sendMetricsToCustomS3(metricSample)
 	case azureProvider:
-		log.Infof("Uploading Metrics to Custom Azure Blob %s", ka.CustomAzureUploadBlobContainerName)
-		go ka.sendMetricsToCustomBlob(ctx, metricSample)
+		if ka.CustomAzureTenantID == "" && ka.CustomAzureClientID == "" {
+			log.Infof("Uploading Metrics to Custom Azure Blob %s with managed identity",
+				ka.CustomAzureUploadBlobContainerName)
+			ka.sendMetricsToCustomBlobManagedIdentity(ctx, metricSample)
```
jdhudson3, line 502: Any reason this is not in a goroutine?

kubernetes/kubernetes.go
```diff
@@ -569,6 +567,20 @@ func (ka KubeAgentConfig) sendMetricsToCustomBlob(ctx context.Context, metricSam
 	ka.uploadBlob(ctx, azureClient, metricSample)
 }
 
+func (ka KubeAgentConfig) sendMetricsToCustomBlobManagedIdentity(ctx context.Context, metricSample *os.File) {
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
+	if err != nil {
+		log.Fatalf("Could not establish Azure credentials with managed identity, "+
```
jdhudson3, line 573: Can we add these two as a startup check for azure as well? If there's an issue it would take 10m for it to fail.

kubernetes/kubernetes.go
AnthonyApptio: changed it to a goroutine

kubernetes/kubernetes.go
AnthonyApptio: added multiple client startup tests for azure (managed identity/service principal) and aws s3

jdhudson3 (approved):

charts/metrics-agent/Chart.yaml
```diff
@@ -14,8 +14,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.11.36
+version: 2.11.37
```
daniel-spray, line 17: You missed one version in values.yaml. It is what makes the helm chart deployment pull in the correct image

AnthonyApptio: thanks i'll update that now

jdhudson3 (dismissed):

PR #279, opened 2024-11-05
Filtering old data out of samples

jdhudson3:
#### What does this PR do?
On especially large clusters, we sometimes see a significant number of old/unused resources such as:
- Jobs that have finished, but are retained
- Pods that are finished (often as a result of the above), but are retained
- Replicasets that have been cycled out, and are not in use

This data is not useful to us when no active pods are running, and there is no reason to add it to the export or process it on the Apptio side. This PR removes the sending of data in these cases.

Unfortunately this cannot be easily filtered out of the informer stored data (would likely require building our own watchers plus other behind the scenes components), but that is a potential future optimization for agent utilization on larger clusters.

Ran on a test cluster, validated sample data did not contain:
- Pods that were terminated
- Replicasets with 0 replicas
- jobs that had completed a while ago

Which were all present in previous samples.

Large clusters tend to orphan more resources, currently we package, upload and store a bunch of stuff that's useless to us.

#### Developer Done List

- [ ] Tests Added/Updated
- [ ] Updated README.md

- [ ] Tests Pass Locally
- [ ] CI Build Passes


retrieval/k8s/k8s_stats.go
```diff
@@ -124,6 +126,38 @@ func writeK8sResourceFile(workDir *os.File, resourceName string,
...
+			canSkip := true
+			for _, v := range resource.Status.ContainerStatuses {
+				if v.State.Terminated != nil && v.State.Terminated.FinishedAt.After(previousHour) {
+					canSkip = false
+				}
+			}
+			return canSkip
+		}
+	case *v1apps.ReplicaSet:
+		if resource.Status.Replicas == 0 && previousHour.After(resource.CreationTimestamp.Time) {
```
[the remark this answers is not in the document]
jdhudson3: Just means that there are no replicas owned by the replicaset, in effect that it is unused and has no pods associated with it.

retrieval/k8s/k8s_stats.go
```diff
@@ -124,6 +126,38 @@ func writeK8sResourceFile(workDir *os.File, resourceName string,
...
+			return true
+		}
+	case *corev1.Pod:
+		if resource.Status.Phase == corev1.PodSucceeded || resource.Status.Phase == corev1.PodFailed {
+			canSkip := true
+			for _, v := range resource.Status.ContainerStatuses {
+				if v.State.Terminated != nil && v.State.Terminated.FinishedAt.After(previousHour) {
+					canSkip = false
+				}
+			}
```
jyin-apptio, line 150: Interest why do we need to this check as Succeeded/Failed are defined below
Succeeded: All containers in the Pod have terminated in success, and will not be restarted.
Failed: All containers in the Pod have terminated, and at least one container has terminated in failure. That is, the container either exited with non-zero status or was terminated by the system, and is not set for automatic restarting.

retrieval/k8s/k8s_stats.go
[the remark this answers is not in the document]
jyin-apptio: Maybe changing to 
```
return resource.Status.Replicas == 0 && previousHour.After(resource.CreationTimestamp.Time)
```
could avoid gocyclo

retrieval/k8s/k8s_stats.go
```diff
@@ -124,6 +126,38 @@ func writeK8sResourceFile(workDir *os.File, resourceName string,
...
+	switch resource := k8Resource.(type) {
+	case *v1batch.Job:
+		if resource.Status.CompletionTime != nil &&
+			previousHour.After(resource.Status.CompletionTime.Time) {
+			return true
+		}
+		if resource.Status.Failed > 0 && resource.Status.StartTime != nil &&
+			previousHour.After(resource.Status.StartTime.Time) {
+			return true
+		}
```
jyin-apptio, line 142: what if this is a long running job and it just failed in this hour? And timestamp for failure time?

retrieval/k8s/k8s_stats.go
```diff
@@ -124,6 +126,38 @@ func writeK8sResourceFile(workDir *os.File, resourceName string,
...
 }
 
+//nolint:gocyclo
+func shouldSkipResource(k8Resource interface{}) bool {
+	// safe buffer to allow for longer lived resources to be ingested correctly
+	previousHour := time.Now().Add(-1 * time.Hour)
+	switch resource := k8Resource.(type) {
+	case *v1batch.Job:
+		if resource.Status.CompletionTime != nil &&
+			previousHour.After(resource.Status.CompletionTime.Time) {
```
jyin-apptio, line 136: Maybe a dumb question, but could the host/status file has different timezone?

jdhudson3: Fair concern, forced UTC

retrieval/k8s/k8s_stats.go
jdhudson3: If it has failed during the current sample period then we wouldn't record it in any case, as the pods will have already stopped. I'm not opposed to bumping it back out further in any case though if we think additional headroom could be valuable.

jyin-apptio: It is not about the `headroom`. I just feel the `startTime` here doesn't make sense.
If your point is we don't record failed pod/job usage, we could ignore it directly. (though I thought we use it and have roadmap to identify crash pods/jobs which consume the resources)

retrieval/k8s/k8s_stats.go
jyin-apptio: > Fair concern, forced UTC

Is k8s timestamp always in UTC?

retrieval/k8s/k8s_stats.go
jdhudson3: My thought for keeping this data was that we may want information related to recently shut down pods.

We could remove all entries and not impact current allocation methodology.

retrieval/k8s/k8s_stats.go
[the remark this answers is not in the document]
jdhudson3: Still too high (15), but seems cleaner

jyin-apptio (dismissed):

jyin-apptio (approved):

PR #288, opened 2025-01-24
Fix Cluster Name Parameter Requirements

Globalgig:
#### What does this PR do?
Adds in some regex to check if the cluster_name config value follows the rules for kubernetes cluster naming.

#### Developer Done List

- [ ] Tests Added/Updated
- [X] Updated README.md

- [ ] Tests Pass Locally
- [ ] CI Build Passes


kubernetes/kubernetes.go
```diff
@@ -123,6 +123,11 @@ const rbacError string = `RBAC role in the Cloudability namespace may need to be
 // CollectKubeMetrics Collects metrics from Kubernetes on a predetermined interval
 // nolint: gocyclo
 func CollectKubeMetrics(config KubeAgentConfig) {
+	isInvalidClusterName, _ := regexp.MatchString(`[^a-zA-Z\d-]`, config.ClusterName)
+	if isInvalidClusterName {
+		log.Fatalf("Invalid agent configuration. CLOUDABILITY_CLUSTER_NAME must contain only alphanumeric characters and dashes "+
+			"and is currently set to %s.", config.ClusterName)
+	}
```
jdhudson3, line 130: Would maybe be more appropriate in `CheckRequiredSettings`

kubernetes/kubernetes.go
```diff
@@ -123,6 +123,11 @@ const rbacError string = `RBAC role in the Cloudability namespace may need to be
 // CollectKubeMetrics Collects metrics from Kubernetes on a predetermined interval
 // nolint: gocyclo
 func CollectKubeMetrics(config KubeAgentConfig) {
+	isInvalidClusterName, _ := regexp.MatchString(`[^a-zA-Z\d-]`, config.ClusterName)
```
jdhudson3, line 126: This check is too specific, currently the only case that can cause problems is a string containing only whitespace.

This would break uploads for many clusters that are otherwise functional.

jdhudson3:
You'll also need to bump the patch version number in a few places.

kubernetes/kubernetes.go
Globalgig: Yep, I mistakenly thought it was supposed to be following the k8s cluster conventions. Changed the logic just to trim whitespace and see if it's an empty string after

kubernetes/kubernetes.go
Globalgig: Moved all the logic over 👍

jdhudson3 (approved):

Globalgig:
Closing because of CLA shenanigans

PR #295, opened 2025-02-28
Reduce payload size

daniel-spray:
#### What does this PR do?
This PR reduces memory usage and payload size of the metrics-agent by stripping unneeded metadata from several Kubernetes resources by default. Previously the metrics-agent had a CLOUDABILITY_PARSE_METRIC_DATA flag that was used to strip potentially sensitive information from k8s resources, this has now been removed and implemented by default for all clusters.

This data stripping was previously done after the resources were collected and stored in memory, now the agent removes unnecessary metadata before storing the resources in the cache (reducing the memory footprint and saving agent running costs! 🎉 ).

Another payload reducing effort is included in this PR where we no longer collect Kubernetes resources that have zero cost associated with them. Such as old Completed Jobs or ReplicaSets that have 0 active pods.

Although every Cluster's impact may vary, the results we see on our test cluster's seems to be about a 40% reduction in payload size.
k8s_stats.go
Deployed to several testing clusters
Nope
#### Developer Done List

- [x] Tests Added/Updated
- [x] Updated README.md

- [ ] Tests Pass Locally
- [ ] CI Build Passes


retrieval/k8s/k8s_stats.go
```diff
@@ -194,36 +245,37 @@ func sanitizeData(to interface{}) interface{} {
...
-		return cast
+		return cast, nil
 	case *corev1.Node:
 		cast := to.(*corev1.Node)
 		sanitizeMeta(&cast.ObjectMeta)
-		return cast
+		return cast, nil
 	}
-	return to
+	return to, fmt.Errorf("invalid object type")
```
jdhudson3, line 271: Any particular reason we want to surface an error from this?

It shouldn't occur as we control both the informers watched and this list of object types. Seems like we'd be introducing an unnecessary risk for minimal value.

kubernetes/kubernetes_test.go
```diff
@@ -415,32 +376,14 @@ func TestCollectMetrics(t *testing.T) {
 				if err != nil {
 					t.Error(err)
 				}
-				if strings.Contains(info.Name(), "pods") {
```
jdhudson3, line 418: Worth keeping some of these asserts around env var removal imo.

retrieval/k8s/k8s_stats.go
```diff
@@ -239,6 +291,12 @@ func sanitizePod(to interface{}) interface{} {
 	for j, container := range (*cast).Spec.InitContainers {
 		(*cast).Spec.InitContainers[j] = sanitizeContainer(container)
 	}
+	(*cast).Spec.DNSPolicy = ""
```
bbylovas, line 294: are we able to completely remove the field from the struct or can we only set it to empty?

daniel-spray: Since it's underlying type is a string, setting it to "" should drop it because the json has a `omitempty` tag.
It is how we sanitize all primitive data types

retrieval/k8s/k8s_stats.go
daniel-spray: I am okay with dropping this, it just seemed a little strange swallowing the error, but it should be fine. If it is none of the types it will just drop the resource anyway

jdhudson3 (approved):

PR #298, opened 2025-03-10
Reduce Metrics Payload Size [2]

daniel-spray:
#### What does this PR do?

(Follow up to https://github.com/cloudability/metrics-agent/pull/295)

This PR reduces memory usage and payload size of the metrics-agent by stripping unneeded metadata from Kubernetes resources by default.

This data stripping was previously done after the resources were collected and stored in memory, now the agent removes unnecessary metadata before storing the resources in the cache (reducing the memory footprint and saving agent running costs! 🎉 ).

Another payload reducing effort is included in this PR where we no longer collect Kubernetes resources that have zero cost associated with them. Such as old Completed Jobs or ReplicaSets that have 0 active pods. @jdhudson3 Thank you!

Although every Cluster's impact may vary, the results we see on our test cluster's seems to be about a 40% reduction in payload size.
k8s_stats.go

Deployed to several testing clusters

Nope
Tested with and without ParseMetricsData enabled
#### Developer Done List

- [x] Tests Added/Updated
- [x] Updated README.md

- [ ] Tests Pass Locally
- [ ] CI Build Passes


retrieval/k8s/k8s_stats.go
```diff
@@ -124,135 +132,189 @@ func writeK8sResourceFile(workDir *os.File, resourceName string,
...
+		if resource.Status.Phase == corev1.PodSucceeded || resource.Status.Phase == corev1.PodFailed {
+			canSkip := true
+			for _, v := range resource.Status.ContainerStatuses {
+				if v.State.Terminated != nil && v.State.Terminated.FinishedAt.After(previousHour) {
+					canSkip = false
+				}
+			}
+			return canSkip
+		}
+	case *v1apps.ReplicaSet:
```
bbylovas, line 164: can we do the same with Deployments?

retrieval/k8s/k8s_stats.go
```diff
@@ -124,135 +132,189 @@ func writeK8sResourceFile(workDir *os.File, resourceName string,
 	return err
 }
 
+//nolint:gocyclo
+func shouldSkipResource(k8Resource interface{}) bool {
```
bbylovas, line 136: love this idea but prob want to run it by Peter first

bbylovas: 100% agreed, i've wanted to add something similar in our pipeline for some time now. mainly just want to make sure all parties are aware

retrieval/k8s/k8s_stats.go
```diff
@@ -124,135 +132,189 @@ func writeK8sResourceFile(workDir *os.File, resourceName string,
...
 		cast := to.(*corev1.Node)
-		sanitizeMeta(&cast.ObjectMeta)
+		sanitizeMeta(&cast.ObjectMeta, parseMetricsData)
 		return cast
 	}
 	return to
 }
 
-func sanitizeMeta(objectMeta *metav1.ObjectMeta) {
+func sanitizeMeta(objectMeta *metav1.ObjectMeta, parseMetricsData bool) {
```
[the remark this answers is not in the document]
daniel-spray: Have a little hesitation to this approach because of the code duplication. Going this route would I think make it more difficult to implement/maintain future sanitization changes as we would need to add it in both methods. Also feel that these are similar enough for now but if you are strong on this separation I can try and implement a clean solution.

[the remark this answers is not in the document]
bbylovas: yeah i think they should be separate and I don't think there shouldn't be any code duplication but maybe I'm missing something so let's sync if so
but main reason for separating both is that although they both pare down the data, they do so for different purposes

1. There's sanitization which is for removing things that customers want to be removed for security purposes or other reasons
2. There's things we drop by default to reduce payload size

there's also the configuration flag that's specifically tied to purpose 1 so it's a bit cleaner to separate that out rather than passing it down through each function and introducing an if-else there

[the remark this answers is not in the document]
daniel-spray: Separated out the two features

retrieval/k8s/k8s_stats.go
```diff
@@ -335,12 +391,18 @@ func sanitizeNamespace(to interface{}) interface{} {
...
+
 func GetTransformFunc(parseMetricsData bool) func(resource interface{}) (interface{}, error) {
 	return func(resource interface{}) (interface{}, error) {
-		var err error
-		resource = sanitizeData(resource, parseMetricsData)
-		if err != nil {
-			return nil, fmt.Errorf("resource is not valid type error: %v", err)
+		if parseMetricsData {
+			resource = sanitizeData(resource)
+		} else {
```
bbylovas, line 406: shouldn't need an `else` block for `trimData`

retrieval/k8s/k8s_stats.go
```diff
@@ -178,154 +177,211 @@ func shouldSkipPod(previousHour time.Time, resource *corev1.Pod) bool {
...
+		(*cast).Spec.Containers[j] = trimContainer(container)
+	}
+	for j, container := range (*cast).Spec.InitContainers {
+		(*cast).Spec.InitContainers[j] = trimContainer(container)
 	}
+	return cast
+}
+
+func sanitizeContainer(container corev1.Container) corev1.Container {
+	container.Env = nil
```
bbylovas, line 372: can remove anything in the `sanitize` funcs that's already in the `trim` funcs

daniel-spray: Updated

bbylovas (approved):

jdhudson3 (approved):

PR #308, opened 2025-06-03
Allow API Key to be Mounted As Volume

daniel-spray:
#### What does this PR do?
Adds support and recommendation to mount the CLOUDABILITY_API_KEY from a k8s secret as a volume instead of reading directly from an environment variable.

Updated a minor version instead of a patch because helm provisioning resources did get altered a bit.
helm chart & kubernetes.go
Deployed on a test cluster using helm command.
Ensured the image can still be deployed using environment variables ensuring backwards compatibility.
We are still planning to support env vars, this is just updating the default method in helm and adding support for volume mounting in the metrics-agent code
#### Developer Done List

- [ ] Tests Added/Updated
- [ ] Updated README.md

- [ ] Tests Pass Locally
- [ ] CI Build Passes

### DIFF
--- a/kubernetes/kubernetes.go
+++ b/kubernetes/kubernetes.go
@@ -247,9 +247,8 @@ func gatherAPIKeyFromVolume(config KubeAgentConfig) string {
 func getKeyFromFileVolume(filepath string) string {
 	key, err := os.ReadFile(filepath)
 	if err != nil {
-		log.Fatalf("error attempting to collect cloudability api key from file: %s with err: %v. "+
-			"Ensure the cloudability api key is either set as an environment variable or the correct volume mounting "+
-			"is in place to allow the agent to access the file storing the api key", filepath, err)
+		log.Warnf("error attempting to collect cloudability api key from file: %s with err: %v", filepath, err)
+		return ""
 	}
 	return string(key)
 }
--- a/kubernetes/kubernetes.go
+++ b/kubernetes/kubernetes.go
@@ -137,9 +137,8 @@ func CollectKubeMetrics(config KubeAgentConfig) {
 	kubeAgent := newKubeAgent(ctx, config)
 
 	customS3Mode := isCustomS3UploadEnvsSet(&kubeAgent)
-
 	if !customS3Mode {
-		kubeAgent.APIKey = gatherAPIKeyFromVolume(config)
+		kubeAgent.APIKey = getAPIKey(config)
 	}
 	// Log start time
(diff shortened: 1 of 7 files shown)


kubernetes/kubernetes.go
```diff
@@ -217,14 +219,38 @@ func CollectKubeMetrics(config KubeAgentConfig) {
 
 }
 
+func gatherAPIKeyFromVolume(config KubeAgentConfig, customS3Mode bool) string {
+	var key string
+	// warning message for agents that are storing the api key as an environment variable
+	if config.APIKey != "" {
+		log.Warnf("warning: cloudability api key is stored as environment variable, best practice is to" +
```
jdhudson3, line 226: Nit / duplicate warn with log level

kubernetes/kubernetes.go
```diff
@@ -217,14 +219,38 @@ func CollectKubeMetrics(config KubeAgentConfig) {
...
+			"use secret store and mount volume, see README for details")
+	}
+	if config.APIKey == "" && !customS3Mode {
```
jdhudson3, line 229: Would prefer flipping these conditions and attempting to load from the file first then read env var since that is the recommended approach now. Can then log separately if both are set, or fail if neither are.

I wouldn't anticipate people having both set, but if/when people switch there's certainly a potential for error/confusion that can be avoided upfront this way.

kubernetes/kubernetes.go
```diff
@@ -217,14 +221,42 @@ func CollectKubeMetrics(config KubeAgentConfig) {
 
 }
 
+func gatherAPIKeyFromVolume(config KubeAgentConfig) string {
```
bbylovas, line 224: nit - would just name this `getAPIKey` since it checks both the vol and env var

kubernetes/kubernetes.go
```diff
@@ -217,14 +221,42 @@ func CollectKubeMetrics(config KubeAgentConfig) {
...
+	}
+	// inform customers who have set both
+	if config.APIKey != "" {
+		log.Warnf("CLOUDABILITY_API_KEY is found in both volume and environment variables." +
+			" Suggest to remove the CLOUDABILITY_API_KEY from environment variables.")
+	}
+	return key
+}
+
+// getKeyFromFileVolume attempts to gather the base64 encoded api key from filepath
```
bbylovas, line 246: qq - does the customer have to encode the key or do we do that for them?

bbylovas (approved):

